PR #3, opened 2019-11-06
implementation of inventory model 0.3

pdowler:


cadc-inventory/src/main/java/org/opencadc/inventory/Artifact.java
```diff
@@ -0,0 +1,189 @@
...
+     * @param o object to compare to
+     * @return true if the logical identifiers are equal, otherwise false
+     */
+    @Override
+    public boolean equals(Object o) {
+        if (o == null) {
+            return false;
+        }
+        Artifact f = (Artifact) o;
+        return uri.equals(f.uri);
```
brianmajor: I remember there being something strange about URI equals() but don't recall what...

cadc-inventory/src/main/java/org/opencadc/inventory/DeletedArtifactEvent.java
```diff
@@ -0,0 +1,89 @@
...
+************************************************************************
+*/
+
+package org.opencadc.inventory;
+
+import java.util.UUID;
+
+/**
+ * Event signifying that an artifact was deleted at a storage site and the deletion
+ * should propagate. The ID of this event if the Artifact ID.
```
brianmajor: if -> is

cadc-inventory/src/main/java/org/opencadc/inventory/DeletedStorageLocationEvent.java
```diff
@@ -0,0 +1,90 @@
...
+*/
+
+package org.opencadc.inventory;
+
+import java.util.UUID;
+
+
+/**
+ * Event signifying that a copy of an artifact was removed/lost at a storage site. 
+ * The ID of this event if the Artifact ID.
```
brianmajor: if -> is

cadc-inventory/src/main/java/org/opencadc/inventory/Entity.java
```diff
@@ -0,0 +1,290 @@
...
+            inc = inc && !Modifier.isStatic(m);
+            // no child entities in storage inventory model
+            //inc = inc && !isChildCollection(f); // 0..* relations to other Entity
+            //inc = inc && !isChildEntity(f); // 0..1 relation to other Entity
+            if (inc) {
+                ret.add(f);
+            }
+        }
+        Class sc = c.getSuperclass();
+        while (sc != null && !Entity.class.equals(sc)) {
```
brianmajor: Not quite understanding this... isn't the recursion enough?  Why the while loop too?

cadc-inventory/src/main/java/org/opencadc/inventory/StorageLocation.java
```diff
@@ -0,0 +1,131 @@
...
+************************************************************************
+*/
+
+package org.opencadc.inventory;
+
+import java.net.URI;
+import java.util.UUID;
+
+/**
+ *
```
brianmajor: Could use a comment explaining the purpose of this class

cadc-inventory/src/main/java/org/opencadc/inventory/StorageSite.java
```diff
@@ -0,0 +1,137 @@
...
+************************************************************************
+*/
+
+package org.opencadc.inventory;
+
+import java.net.URI;
+import org.apache.log4j.Logger;
+
+/**
+ *
```
brianmajor: Could use a comment explaining the purpose of this class

cadc-inventory/src/main/java/org/opencadc/inventory/Entity.java
pdowler: Recursion os for child objects in the tree (not technically used in this model); this traversal is to find all member fields in this object declared here or in super classes. While that's also not used in this model I kept the method as-is because I intend to pull the common parts out of this and caom2 libs to someplace.

cadc-inventory/src/main/java/org/opencadc/inventory/Artifact.java
pdowler: UUID.compareTo in java doesn't comply with the UUID spec. Maybe that was it?

cadc-inventory/src/main/java/org/opencadc/inventory/Entity.java
```diff
@@ -91,13 +91,22 @@
...
+        this(null);
+    }
+    
+    protected Entity(UUID id) {
+        if (id == null) {
+            this.id = UUID.randomUUID();
+        } else {
+            this.id = id;
+        }
     }
```
brianmajor: Maybe just make the no-arg contructor public?  Then you could check for accidentally null UUIDs in the constructor with a parameter.

cadc-inventory/src/main/java/org/opencadc/inventory/Artifact.java
brianmajor: Hmm, maybe.  Javadoc for URI.equals() seems like it's doing the right thing anyhow.

cadc-inventory/src/main/java/org/opencadc/inventory/Entity.java
pdowler: simplified constructors; they are protected because technically you don't instantiate a bare Entity... making it abstract

PR #13, opened 2019-11-29
add StorageLocation.storageBucket

pdowler:
separates Artifact.uriBucket from StorageLocation.storageBucket (eg AD can set the archive name as the storageBucket)

- updated UML and VO-DML
- updated cadc-inventory
- updated cadc-inventory-db


cadc-inventory-db/src/main/resources/inventory.Artifact.sql
```diff
@@ -9,7 +9,8 @@ create table <schema>.Artifact (
     contentEncoding varchar(128),
 
     siteLocations uuid[],
-    storageLocation varchar(256),
+    storageLocation_storageID varchar(256),
+    storageLocation_storageBucket varchar(8);
```
brianmajor: Do you think 8 is long enough?  Perhaps it should be 32 to support UUIDs?

pdowler: For the bucket you want plain strings so you can support a shorter prefix (smaller number of larger buckets). 5 chars of hex is 1 million buckets. AD would use the archie name here, which fits into 8.

pdowler: Of course, we can trivially make it longer later if needed.

pdowler:
Artifact.uriBucket still allows for metadata-validator to batch; since it will query via the TAP API (to the remote) it will be able to stream and batch. Aiming at support for 1 billion artifacts so thousands of buckets with up to 1 million files each (caom2 validation of HST is ~2 million observations: takes 3-4 minutes; 4GiB ram because it doesn't stream/merge and holds two lists in memory).

cadc-inventory-db/src/main/resources/inventory.Artifact.sql
brianmajor: Presumably the storageBucket is for non-standard use (like with ad) so doesn't necessarily support bucket prefixes.  But yes, we can always extend it.

PR #15, opened 2019-12-03
File system storage adapter

brianmajor:
Looking at the content in build/tmp/fsroot after building the project gives an idea of the two bucket modes that are supported.


cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/ArtifactIterator.java
```diff
@@ -0,0 +1,123 @@
...
+        URI storageID = meta.getStorageLocation().getStorageID();
+        String ssp = storageID.getSchemeSpecificPart();
+        int firstSlash = ssp.indexOf("/");
+        if (firstSlash < 1) {
+            log.debug("unrecognized storageID format");
+        } else {
+            StringBuilder sb = new StringBuilder();
+            sb.append(ssp.substring(0, firstSlash));
+            sb.append(ssp.substring(firstSlash + 1));
+            meta.artifactURI = URI.create(sb.toString());
```
pdowler: Presumably if URI.create throws an exception it is a bug in this library... should probably use "new URI" and then

```
catch (URISyntaxException ex) {
    throw new RuntimeException("BUG: failed to reconstruct Artifact.uri for " + meta, ex);
}
```

cadc-storage-adapter/src/main/java/org/opencadc/inventory/storage/StorageAdapter.java
```diff
@@ -122,6 +122,9 @@ public void get(StorageLocation storageLocation, InputStreamWrapper wrapper, Set
      * Write an artifact to storage.
      * The value of storageBucket in the returned StorageMetadata and StorageLocation can be used to
      * retrieve batches of artifacts in some of the iterator signatures defined in this interface.
+     * The bucket value in the supplied artifact may be used by the storage implementation
+     * to organize files.  Batches of artifacts can be listed by bucket in two of the
+     * iterator signatures in this interface.
```
pdowler: Should clarify that if storageBucket is null in the returned StorageLocation, then the caller will never be able to do batched (bucket-based) validation via the iterator(String) method.

The comment that "Artifact.uriBucket may be used" should come with the caveat that rename will change that bucket value (assuming we stick with uriBucket instead of pkBucket - TBD)... which implies that the StorageAdapter.rename(...) method probably needs to be specified clearly to go with any mention of the artifact bucket scheme.

brianmajor: Okay.  I'm going to leave it as is until we decide what ID uriBucket is based on.  The implementation is currently assuming pkBucket.

pdowler:
You might as well delete the .travis.yml file - I'm not likely to try to make it work now that github actions do work.

brianmajor:
Yes, please move the computeBucket method to InventoryUtil.  Perhaps a note should accompany it saying that changing the implementation can result in inconsistent storage in some filesystems.

brianmajor:
done

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -0,0 +1,523 @@
...
+                    + "but calculated [" + length + "]");
+            }
+        } else {
+            log.debug("No contentLength provided.");
+        }
+        
+        URI storageID = null;
+        String storageBucket = null;
+        switch (bucketMode) {
+            case URI_BASED:
```
pdowler: Part of the URI_BASED case is the same sequence of appends as in createArtifactPath. That might be easier to not duplicate if createArtifactPath was createStorageLocation and returned a StorageLocation. Then the same code as used bu get() can be used to actually get the Path to write to.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -0,0 +1,523 @@
...
+                for (char c : uriBucket.toCharArray()) {
+                    path.append(c).append(File.separator);
+                }
+                UUID filename = UUID.randomUUID();
+                path.append(filename);
+                break;
+            default:
+                throw new IllegalStateException("unsupported bucket mode");
+        }
+        return root.resolve(path.toString());
```
pdowler: I wonder if this method should have created and returned a StorageLocation? The put(...) is parsing the path returned here to create the StorageLocation.  You need StorageLocation-to-Path code for get(...) anyway so put() could use that to figure out where to write... 

It seems like all the separate bits are here and they get serialised and re-parsed again.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -0,0 +1,523 @@
...
+        if (!Files.exists(ret)) {
+            throw new ResourceNotFoundException("not found: " + storageID);
+        }
+        if (!Files.isRegularFile(ret)) {
+            throw new IllegalArgumentException("not found: " + storageID);
+        }
+        return ret;
+    }
+    
+    public static URI createMD5Checksum(Path path) throws NoSuchAlgorithmException, IOException {
```
pdowler: What is the purpose of this method?

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -0,0 +1,523 @@
...
+                if (throwable instanceof IOException) {
+                    throw (IOException) throwable;
+                }
+                // TODO: identify throwables that are transient
+                throw new IllegalStateException("Unexpected error", throwable);
+            }
+        }
+        
+        // checksum comparison
+        if (newArtifact.contentChecksum != null && newArtifact.contentChecksum.getScheme().equals(MD5_CHECKSUM_SCHEME)) {
```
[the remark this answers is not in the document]
brianmajor: good call

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -0,0 +1,523 @@
...
+            if (!expectedLength.equals(length)) {
+                throw new StreamCorruptedException(
+                    "expected contentLength [" + expectedLength + "] "
+                    + "but calculated [" + length + "]");
+            }
+        } else {
+            log.debug("No contentLength provided.");
+        }
+        
+        URI storageID = null;
```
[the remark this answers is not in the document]
brianmajor: okay, I agree.  It's all in the try block now.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -0,0 +1,523 @@
...
+    /**
+     * Delete from storage the artifact identified by storageID.
+     * @param storageLocation Identifies the artifact to delete.
+     * 
+     * @throws ResourceNotFoundException If the artifact could not be found.
+     * @throws IOException If an unrecoverable error occurred.
+     * @throws TransientException If an unexpected, temporary exception occurred. 
+     */
+    public void delete(StorageLocation storageLocation) throws ResourceNotFoundException, IOException, TransientException {
+        InventoryUtil.assertNotNull(FileSystemStorageAdapter.class, "storageLocation", storageLocation);
```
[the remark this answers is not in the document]
brianmajor: It's there already

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -0,0 +1,523 @@
...
+            return iterator;
+        } catch (InvalidPathException e) {
+            throw new IllegalArgumentException("Invalid bucket: " + bucket);
+        }
+    }
+    
+    private Path createArtifactPath(URI artifactURI, String uriBucket) {
+        StringBuilder path = new StringBuilder();
+        switch (bucketMode) {
+            case URI_BASED:
```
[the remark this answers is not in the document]
brianmajor: Yes, the directory structure in this mode for an artifactURI of cadc:TEST/my/file would be fsroot/cadc:/TEST/my/file.  Bucketing in this mode can be any depth of directory, so you could bucket by cadc: or cadc:TEST or cadc:TEST/my.

However, note how there are no slashes after the scheme in these bucket examples.  I decided to let the bucket resemble the artifactURI substrings rather than the storageURI substrings for usability.  So, cadc:TEST rather than cadc:/TEST.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
brianmajor: It's used by the file system iterator.  I'm not sure why it's in this class though.  I'll move it.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
brianmajor: Took me a while to understand this suggestion but got it now.  I've refactored put so it creates the storage location upfront and uses the existing method to get the path from the location.  It fell out nicely.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
brianmajor: Yep, see reply above.

PR #17, opened 2019-12-05
Initial commit of cadc-storage-permissions.

jburke-cadc:


cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadWriteGrant.java
```diff
@@ -0,0 +1,109 @@
...
+        return groups;
+    }
+
+    /**
+     * Get date after which the Grant is considered expired and should be renewed.
+     * @return The grant expiry date.
+     */
+    public Date getExpiryDate() {
+        return expiryDate;
+    }
```
brianmajor: Why is getExpiryDate() not just implemented in Grant instead of being abstract?

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadGrant.java
```diff
@@ -0,0 +1,120 @@
...
+        return expiryDate;
+    }
+
+    /**
+     * Check if artifact access is available to anonymous (all) users.
+     * @return true if the artifact has anonymous access, false otherwise.
+     */
+    public boolean isAnonymousAccess() {
+        return anonymousAccess;
+    }
```
brianmajor: How does one set anonymous access to true?  I guess you need to add 
 an anonymousAccess param to the constructor if it's required.

cadc-storage-permissions/src/main/webapp/capabilities.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<vosi:capabilities
+    xmlns:vosi="http://www.ivoa.net/xml/VOSICapabilities/v1.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns:vs="http://www.ivoa.net/xml/VODataService/v1.1">
+
+  <capability standardID="vos://cadc.nrc.ca~vospace/CADC/std/inventory/#TBD-1.0">
```
brianmajor: I chose vos://cadc.nrc.ca\~vospace/CADC/std/inventory#artifacts-1.0 for minoc, but it could perhaps be #files-1.0 instead.
This one could be vos://cadc.nrc.ca\~vospace/CADC/std/inventory#permissions-1.0

jburke-cadc:
> I'm guessing the Grant object was split because anonymous access doesn't apply to write permission? If so then shouldn't it be called just WriteGrant, instead of ReadWriteGrant? (and the client has methods getReadGrant and getWriteGrant)

Yes. Splitting into read and write methods (agree read is redundant in read-write) is to minimize the work needed to get either set of permissions. I will rename the ReadWriteGrant method.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/Grant.java
```diff
@@ -83,20 +82,15 @@
...
     private final URI artifactURI;
-    
-    public Date releaseDate;
-    public boolean isPublic = false;
-    public final Date evaluationDate = new Date();
-    
-    private final List<GroupURI> readGroups = new ArrayList<GroupURI>();
-    private final List<GroupURI> readWriteGroups = new ArrayList<GroupURI>();
+    protected final List<GroupURI> groups = new ArrayList<GroupURI>();
+    protected final Date expiryDate = new Date();
```
pdowler: expireDate cannot be final and set to "now"... different sources of permissions will decide on the expiryDate. If it is mandatory (I think it should be) then the field should be private and it should be included in the constructor args.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/Grant.java
```diff
@@ -105,26 +99,22 @@ public Grant(URI artifactURI) {
...
-     * Get the read group list.
-     * @return The read groups
+     * Get date after which the Grant is considered expired and should be renewed.
+     * @return The grant expiry date.
      */
-    public List<GroupURI> getReadGroups() {
-        return readGroups;
-    }
-
+    public abstract Date getExpiryDate();
```
pdowler: getExpiryDate() is required but it doesn't need to be abstract; just return the value.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/Grant.java
```diff
@@ -105,26 +99,22 @@ public Grant(URI artifactURI) {
...
     /**
-     * Get the read-write group list.
-     * @return The read-write groups
+     * Get the group list.
+     * @return The list of groups.
      */
-    public List<GroupURI> getReadWriteGroups() {
-        return readWriteGroups;
-    }
+    public abstract List<GroupURI> getGroups();
```
pdowler: Not abstract. Just normal return.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadWriteGrant.java
jburke-cadc: I wasn't sure how it is determined? If it's a set period of time then it could be in Grant. Or is there another way to calculate and set it based on the archive when getting the permissions? Is it the same for read and write?

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/PermissionsClient.java
```diff
@@ -89,17 +88,33 @@ public PermissionsClient() {
...
+    /**
+     * Get the read-write permissions information about the file identified by fileURI.
+     *
+     * @param artifactURI Identifies the artifact for which to retrieve grant information.
+     * @return The read-write grant information.
+     *
+     * @throws ResourceNotFoundException If the file could not be found.
+     * @throws TransientException If an unexpected, temporary exception occurred.
+     */
+    public ReadWriteGrant getReadWriteGrant(URI artifactURI)
```
pdowler: Agree with other review that this should, be getWriteGrant. The read-only vs read-write is a concept in the way we maintain permissions (read-write means "this group has two permissions")... but here these methods are about what action is being authlorised.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadGrant.java
jburke-cadc: At some point when getting the read permissions would there be a point where you can say 'the artifact is anonymous access, set the flag true, and return' was my thought?

cadc-storage-permissions/src/main/webapp/capabilities.xml
jburke-cadc: permissions makes sense. We haven't named the project yet, hopefully tomorrow since I'm getting tired of typing cadc-storage-permissions, and was going to raise the standardID name then. Or just change it to permissions now.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/Grant.java
jburke-cadc: Agree it should be mandatory, can always be a large value if it's not applicable for some reason. Added to the constructor.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/Grant.java
jburke-cadc: Done.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/Grant.java
jburke-cadc: Also done.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/PermissionsClient.java
jburke-cadc: Renamed.

jburke-cadc:
> cadc-storage-permissions is a library with client and "payload" classes. Will eventually have Reader and Writer classes. It is not a docker build. ServiceAvailability and all the webapp stuff and war plugin don't belong here... they will go into a module named TBD.
> 
> Until more code is written, the get*Grant(URI) methods should throw UnsupportedOperationException

I was looking at this from the wrong direction. I thought this was the app, hence the renaming, and the API would be sorted out and then extracted into a library. Removed the non-library files and created a separate module for the app.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadGrant.java
brianmajor: Yes, but I mean there is no way to set it to true in the object, is there?

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadWriteGrant.java
brianmajor: I don't know what concept it is trying to portray.  I originally had an 'evaluation date' field where my intent was that clients could see this date to determine if the information was recent enough to be considered correct.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/PermissionsClient.java
```diff
@@ -89,17 +88,33 @@ public PermissionsClient() {
...
+    public ReadGrant getReadGrant(URI artifactURI)
+        throws ResourceNotFoundException, TransientException {
+
+        throw new UnsupportedOperationException("Not implemented");
     }
-    
+
+    /**
+     * Get the read-write permissions information about the file identified by fileURI.
+     *
```
brianmajor: change comment here to say 'write', not 'read-write'

brianmajor: Also, you still need to provide a way to set annonymousAccess.  If it's mandatory it should be a param in the constructor...

jburke-cadc: Changed and setAnonymousAccess method added.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadGrant.java
jburke-cadc: Method added to set anonymous access.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadWriteGrant.java
jburke-cadc: Same idea but a more explict name to indicate the service doesn't consider the grant reliable past that time and it should be renewed. Saying the grant has now `expired` is a strong hint to renew it.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadGrant.java
brianmajor: No setters in these model objects... you can just make anonymousAccess a public attribute and remove the getter and setter.

jburke-cadc: Agree if it's mandatory it should be in the constructor, added.

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/ReadGrant.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Construct a read grant for the given artifactURI and expiry date.
+     * @param artifactURI The applicable targetURI.
+     * @param expiryDate The expiry date of the grant.
+     * @param anonymousAccess true is the artifact has anonymous access, false otherwise.
+     */
+    public ReadGrant(URI artifactURI, Date expiryDate, boolean anonymousAccess) {
+        super(artifactURI, expiryDate);
+        InventoryUtil.assertNotNull(ReadGrant.class, "anonymousAccess", anonymousAccess);
```
brianmajor: You can remove this check because primitives can't be null.

brianmajor: Just the one change above... otherwise looks good.

jburke-cadc: Done

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/PermissionsClient.java
```diff
@@ -89,17 +88,33 @@ public PermissionsClient() {
...
+    /**
+     * Get the write permissions information about the file identified by fileURI.
+     *
+     * @param artifactURI Identifies the artifact for which to retrieve grant information.
+     * @return The read-write grant information.
+     *
+     * @throws ResourceNotFoundException If the file could not be found.
+     * @throws TransientException If an unexpected, temporary exception occurred.
+     */
+    public WriteGrant getReadWriteGrant(URI artifactURI)
```
pdowler: This should be getWriteGrant(URI)

cadc-storage-permissions/build.gradle
```diff
@@ -22,8 +22,9 @@ dependencies {
     compile 'org.opencadc:cadc-inventory:[0.1,)'
     compile 'org.opencadc:cadc-gms:[1.0,)'
 
+    runtime 'org.opencadc:cadc-log:[1.1,)'
```
pdowler: cadc-log not needed

cadc-storage-permissions/src/main/java/org/opencadc/inventory/permissions/PermissionsClient.java
jburke-cadc: Updated.

cadc-storage-permissions/build.gradle
jburke-cadc: Removed.

jburke-cadc:
> cadc-storage-permissions is a library with client and "payload" classes. Will eventually have Reader and Writer classes. It is not a docker build. ServiceAvailability and all the webapp stuff and war plugin don't belong here... they will go into a module named TBD.
> 
> Until more code is written, the get*Grant(URI) methods should throw UnsupportedOperationException

Done

PR #18, opened 2019-12-05
storage adapter api changes - no stream wrappers

brianmajor:
- Only InputStream and OutputSteam.
- Three new checked exceptions.
- Moved IO support to separate class ThreadedIO
- Tests in storage-adapter and storage-adapter-fs use ThreadedIO directly


cadc-storage-adapter/src/main/java/org/opencadc/inventory/storage/StorageClient.java
```diff
@@ -94,59 +90,38 @@
  * @author majorb
  *
  */
-public class StorageClient {
+public class StorageClient implements StorageAdapter {
```
pdowler: Does this class do anything beyond loading the StorageAdapter impl? It should probably just be replaced with something like a factory class that reads config and instantiates the impl.

cadc-storage-adapter/src/main/java/org/opencadc/inventory/storage/StorageEngageException.java
```diff
@@ -0,0 +1,96 @@
...
+package org.opencadc.inventory.storage;
+
+/**
+ * This exception indicates that the storage adapter failed
+ * to connect to or interact with its storage system.
+ * 
+ * @author majorb
+ */
+@SuppressWarnings("serial")
+public class StorageEngageException extends Exception {
```
pdowler: Not crazy about this name but don't have any better idea :-)

cadc-storage-adapter/src/main/java/org/opencadc/inventory/storage/StorageClient.java
pdowler: Ultimately, minoc's ArtifactAction will have to instantiate the impl and make it available to subclasses. That could be a single method there rather than another class... and probably /conf/minoc.properties which is read on startup (since you can't change the adapter impl at runtime) but different instances will need different impls.

brianmajor: No, it's not very useful any longer.  I'll get rid of it.

cadc-storage-adapter/src/main/java/org/opencadc/inventory/storage/StorageEngageException.java
brianmajor: Me neither

PR #24, opened 2019-12-10
minoc core functionality

brianmajor:


minoc/src/main/java/org/opencadc/minoc/HeadAction.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Download the artifact or cutouts of the artifact.
+     * @param artifactURI The identifier for the artifact. 
+     * @return The artifact
+     */
+    @Override
+    public Artifact execute(URI artifactURI) throws Exception {
+        ArtifactDAO dao = getArtifactDAO();
+        Artifact artifact = getArtifact(artifactURI, dao);
+
```
pdowler: Making all these temp variables is pointless and makes writing clear code below harder. Just use the get methods or fields directly. For those that can be accessed via a field, setHeader should only be called when the value is not null. That's going to be more clear without local variables.

minoc/src/main/java/org/opencadc/minoc/GetAction.java
```diff
@@ -90,11 +92,18 @@ public GetAction() {
...
      * Download the artifact or cutouts of the artifact.
-     * @param artifactURI The identifier for the artifact. 
+     * @param artifactURI The identifier for the artifact.
+     * @return The artifact
      */
     @Override
-    public void execute(URI artifactURI) throws Exception {
-        // TODO
+    public Artifact execute(URI artifactURI) throws Exception {
+        Artifact artifact = super.execute(artifactURI);
```
pdowler: This super call to HeadAction to re-use the setHeader calls seems more complex/subtle than just having a static method like HeadAction.writeHeaders(Artifact, SyncOutput) that both doAction methods call. I missed this fact the fuirst read through and was going to comment that you need to set headers because nothing here says "set headers".

minoc/src/main/java/org/opencadc/minoc/PostAction.java
```diff
@@ -0,0 +1,133 @@
...
+        String newContentEncoding = syncInput.getParameter("contentEncoding");
+        log.debug("new uri: " + newURI);
+        log.debug("new contentType: " + newContentType);
+        log.debug("new contentEncoding: " + newContentEncoding);
+        
+        ArtifactDAO dao = getArtifactDAO();
+        Artifact artifact = getArtifact(artifactURI, dao);
+        
+        // TODO: enable modifying URIs when supported by DAO
+        // TODO: how to support clearing values?
```
pdowler: You don't need to make comp0lex logic here to see if something changed. The ArtifactDAO recomputes accMetaChecksum and astModified and only writes to the database if there are changes already.

Not sure about clearing values... only applicable to contentEncoding

minoc/src/main/java/org/opencadc/minoc/PutAction.java
```diff
@@ -0,0 +1,202 @@
...
+        artifact.storageLocation = artifactMetadata.getStorageLocation();
+        
+        ArtifactDAO dao = getArtifactDAO();
+        Artifact existing = dao.get(artifactURI);
+        if (existing == null) {
+            dao.put(artifact);
+            log.debug("put artifact in database: " + artifactURI);
+        } else {
+            try {
+                dao.getTransactionManager().startTransaction();
```
pdowler: You need to create and put a DeletedArtifactEvent

minoc/src/main/java/org/opencadc/minoc/PutAction.java
```diff
@@ -0,0 +1,202 @@
...
+            } catch (Throwable t) {
+                log.error("failure replacing artifact, attempting rollback");
+                try {
+                    dao.getTransactionManager().rollbackTransaction();
+                    log.error("rollback successful");
+                } catch (Exception e) {
+                    log.error("failed to rollback transaction", e);
+                }
+                throw t;
+            }
```
pdowler: MUST have a finally that detects an open transaction and calls rollback

minoc/src/main/java/org/opencadc/minoc/DeleteAction.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public Artifact execute(URI artifactURI) throws Exception {
+        ArtifactDAO dao = getArtifactDAO();
+        Artifact artifact = getArtifact(artifactURI, dao);
+        
+        dao.delete(artifact.getID());
+        log.debug("deleting from storage...");
+        getStorageAdapter().delete(artifact.storageLocation);
+        log.debug("deletedfrom storage");
+
```
pdowler: When you delete an artifact, you have to also create a DeletedArtifactEvent and put it with the DeletedEventDAO. You need to do both in a transaction. The second DAO is instantiated using the copy constructor so they share config, data source, and transaction manager. 

So there are really 3 actions here and only 2 can be made atomic with a database transaction. This is the crux of the distributed transaction we discussed and the decision about ordering: what if the second action fails? What error does the service report to the user?

1. If you delete from inventory and then fail, the next file-validate could recover the Artifact (if the StorageAdapter impl returns artifactURI in the iterator). You would report that the delete failed, but the artifact would be gone for awhile, that would get propagated, the artifact would be recovered, that would get propagated... so that's ultimately a failure to delete that looks like it worked fior awhile. ughh.

2. If you delete from storage first and then fail, you have an artifact in iventory but no file in storage: later GETs will fail. You would report that the delete failed to the caller, but the artifact would still be there and in future you may recover or complete the deletion. The next file-validate could possibly schedule a new download (it would set the storageLocation to null so that file-sync would pick it up) but only if another site has a copy already... or it would delete the artifact if there are no other copies and at that time the DeletedArtifactEvent would propagate; until then, global transfer negotiation would keep handing out URLs that don't work.

For now, if you delete from inventory and fire event first (in transaction), delete from storage while txn is open, and rollback if anything fails we end up with a very small chance of scenario 2.

minoc/Dockerfile
```diff
@@ -3,4 +3,9 @@
 
 FROM cadc-tomcat:latest
 
+RUN ln -s /conf /root/config
+
+RUN rm -rf /usr/share/tomcat/webapps/*
 COPY build/libs/*.war /usr/share/tomcat/webapps
+
+EXPOSE 8080
```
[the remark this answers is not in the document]
brianmajor: As far as I can tell EXPOSE is for documentation purposes and does not actually change the image at all.

minoc/src/main/java/org/opencadc/minoc/ArtifactUtil.java
```diff
@@ -102,7 +102,7 @@
      * Valid methods that apply to authorization.
      */
     public static enum HttpMethod {
-        GET, PUT, POST, DELETE
+        GET, PUT, POST, DELETE, HEAD
```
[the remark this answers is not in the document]
brianmajor: yes, good ideas.

minoc/Dockerfile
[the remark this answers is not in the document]
brianmajor: I changed */war to minoc.war but it's not showing up in this diff for some reason.

minoc/src/main/java/org/opencadc/minoc/ArtifactAction.java
```diff
@@ -100,6 +99,10 @@
 public abstract class ArtifactAction extends RestAction {
     private static final Logger log = Logger.getLogger(ArtifactAction.class);
     
+    public static final String JNDI_DATASOURCE = "jdbc/inventory";
+    public static final String DATABASE = "content";
+    public static final String SCHEMA = "inventory";
+
```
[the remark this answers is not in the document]
brianmajor: done

minoc/src/main/java/org/opencadc/minoc/ArtifactAction.java
```diff
@@ -129,9 +139,10 @@ protected InlineContentHandler getInlineContentHandler() {
      * Do the work of the subclass.
      * 
      * @param artifactURI The target artifact
+     * @return The artifact (if applicable)
      * @throws Exception If an something goes wrong.
      */
-    public abstract void execute(URI artifactURI) throws Exception;
+    public abstract Artifact execute(URI artifactURI) throws Exception;
```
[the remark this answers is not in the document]
brianmajor: all refactored

minoc/Dockerfile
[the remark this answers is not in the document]
brianmajor: okay, there now... had to refresh this view

minoc/src/main/java/org/opencadc/minoc/DeleteAction.java
brianmajor: done

minoc/src/main/java/org/opencadc/minoc/ArtifactAction.java
```diff
@@ -191,21 +202,103 @@ void parsePath() {
...
+        if (deletedEventDAO == null) {
+            DeletedEventDAO newDAO = new DeletedEventDAO(src);
+            deletedEventDAO = newDAO;
+        }
+        return deletedEventDAO;
+    }
+    
+    static Map<String, Object> getDaoConfig() {
+        Map<String, Object> config = new HashMap<String, Object>();
+        PropertiesReader pr = new PropertiesReader("minoc.properties");
```
pdowler: The minoc.properties file is read twice -- should read once (maybe lazy init of a "PropertiesReader props" member variable?)

pdowler: And document the required props in catalina.properties to configure the pool in context.xml

brianmajor: minoc.properties is read once per request (whether it's from availability or one of the actions).  The other dao just uses the copy constructor.

pdowler: I wasn't clear - it is read once to get the StorageAdapter impl  and once to get the SQLGenerator impl and database info.

On minoc.properties:

Technically, knowing the database name is not required since in most cases is it in the JDBC url (catalina.properties). Putting it into the InitDatabase (ctor) means it is used to qualify the table name... SQLGenerator currently ignores the database setting and I'll only add support for putting it into the SQL if needed. In most cases: configuring it in two places only adds an opportunity for a mistake. So that should probably be left out for now and a null database name passed into the init. I'll probably change InitDatabase to ignore the database until both it and SQLGenerator need it.

brianmajor: Re: reading config--since this pull request I've changed it so that the MultiValuedProperties are obtained once from the file and used from memory to get all config items.

Okay, I'll remove database and make SQL generator optional.

PR #28, opened 2019-12-17
completion of minoc core

brianmajor:


minoc/src/intTest/java/org/opencadc/minoc/MinocIntTest.java
```diff
@@ -116,84 +115,16 @@ public MinocIntTest() {
         RegistryClient regClient = new RegistryClient();
         anonURL = regClient.getServiceURL(MINOC_SERVICE_ID, MINOC_STANDARD_ID, AuthMethod.ANON);
         log.info("anonURL: " + anonURL);
+        certURL = regClient.getServiceURL(MINOC_SERVICE_ID, MINOC_STANDARD_ID, AuthMethod.CERT);
+        log.info("certURL: " + certURL);
+        anonSubject = AuthenticationUtil.getAnonSubject();
+        File cert = FileUtil.getFileFromResource("x509_CADCRegtest1.pem", MinocIntTest.class);
```
pdowler: We should hard code a neutral name like minoc-test.pem here; developers will have to make a symlink or copy of the cert they want to use.

minoc/src/intTest/java/org/opencadc/minoc/VosiCapabilitiesTest.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+    
+    public VosiCapabilitiesTest() {
+        super(BasicOpsIntTest.MINOC_SERVICE_ID);
+    }
+
+    @Override
+    protected void validateContent(Capabilities caps) throws Exception {
+        super.validateContent(caps);        
+        // TODO: add checking for the minoc capabilities as per the example below
```
pdowler: Should be able to assert some content now

minoc/src/main/webapp/WEB-INF/web.xml
```diff
@@ -13,7 +13,7 @@
     <servlet-class>ca.nrc.cadc.log.LogControlServlet</servlet-class>
     <init-param>
       <param-name>logLevel</param-name>
-      <param-value>info</param-value>
+      <param-value>debug</param-value>
```
pdowler: info

PR #65, opened 2020-02-21
Adding profiler information for further test runs.

at88mph:


minoc/src/main/java/org/opencadc/minoc/PutAction.java
```diff
@@ -164,21 +165,27 @@ public void doAction() throws Exception {
...
-        
+
+        putProfiler.checkpoint("getContent(inputstream)");
+
         StorageMetadata artifactMetadata = null;
         
         log.debug("writing new artifact to storage...");
         try {
             artifactMetadata = getStorageAdapter().put(newArtifact, in);
+            putProfiler.checkpoint("getStorageAdapter().put() success.");
```
pdowler: Would be good to separate the getStorageAdapter() call from put() and add an extra profiler call between them, which could be non-zero in some implementations.

storageAdapter.init
storageAdapter.put.ok

minoc/src/main/java/org/opencadc/minoc/PutAction.java
```diff
@@ -189,10 +196,12 @@ public void doAction() throws Exception {
         artifact.contentEncoding = encodingHeader;
         artifact.contentType = typeHeader;
         artifact.storageLocation = artifactMetadata.getStorageLocation();
-        
+
+        final Profiler databaseProfiler = new Profiler(PutAction.class);
```
pdowler: Don't need a new profiler here, just keep using the first one

minoc/src/main/java/org/opencadc/minoc/PutAction.java
```diff
@@ -189,10 +196,12 @@ public void doAction() throws Exception {
...
         ArtifactDAO artifactDAO = getArtifactDAO();
         ObsoleteStorageLocationDAO locDAO = new ObsoleteStorageLocationDAO(artifactDAO);
         Artifact existing = artifactDAO.get(artifactURI);
+        databaseProfiler.checkpoint(String.format("artifactDAO.get(%s)", artifactURI));
```
pdowler: artifactDAO.get

minoc/src/main/java/org/opencadc/minoc/PutAction.java
```diff
@@ -164,21 +165,27 @@ public void doAction() throws Exception {
...
             artifactMetadata = getStorageAdapter().put(newArtifact, in);
+            putProfiler.checkpoint("getStorageAdapter().put() success.");
         } catch (ReadException e) {
             // error on client read
             String msg = "read input error";
             log.debug(msg, e);
             if (e.getMessage() != null) {
                 msg += ": " + e.getMessage();
             }
+            putProfiler.checkpoint(String.format("getStorageAdapter().put() failure \n%s.", msg));
```
pdowler: storageAdapter.put.fail

PR #68, opened 2020-02-25
Initial README, Dockerfile, and some infrastructure for cadc-file-validate.

at88mph:
Trying to issue smaller reviews at a time to avoid a big one at the end.


cadc-file-validate/README.md
```diff
@@ -0,0 +1,74 @@
+# cadc-file-validate 0.1
+
```
pdowler: In the model this component is simply denoted as a _process_ without saying exactly how it is implemented... some of them may run and terminate like a command line tool and others may run forever and manage their own idle time... better to call it a something vague like _process_ or _tool_ but not command-line.

cadc-file-validate/README.md
```diff
@@ -0,0 +1,74 @@
...
+Command line tool to ensure validity of the information stored in the Inventory Database and what is served from a Storage Adapter.  
+
+This is intended to run periodically at any site to determine what changes, if any, are to be made to either the Storage Inventory database, or
+the storage system.
+
+## Configuration
+
+A file called `cadc-file-validate.properties` should be available from the `/config` directory.  All properties in the file are loaded and set
+as System properties:
+
```
pdowler: Don't need to show java sample code.

cadc-file-validate/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+## Checking it
+```
+docker run -t opencadc/cadc-file-validate:latest
+```
+
+### With usage
+```
+docker run -t opencadc/cadc-file-validate:latest cadc-file-validate -h
```
pdowler: -h?

cadc-file-validate/README.md
```diff
@@ -0,0 +1,74 @@
...
+#org.opencadc.inventory.storage.StorageAdapter={fully-qualified-classname of implementation}
+org.opencadc.inventory.storage.StorageAdapter=org.opencadc.inventory.storage.fs.FileSystemStorageAdapter
+
+# The SQL generator implementation (default shown)
+#org.opencadc.inventory.db.SQLGenerator=org.opencadc.inventory.db.SQLGenerator
+
+# The schema to use
+org.opencadc.inventory.db.schema=inventory
+
+# The identity of a permissions service(s) providing read permissions.
```
pdowler: permissions services not applicable

cadc-file-validate/README.md
```diff
@@ -0,0 +1,74 @@
...
+# Consult the storage adapter instructions for adapter-specific configuration.  The library MUST be
+# accessible from this application's classpath.
+#org.opencadc.inventory.storage.StorageAdapter={fully-qualified-classname of implementation}
+org.opencadc.inventory.storage.StorageAdapter=org.opencadc.inventory.storage.fs.FileSystemStorageAdapter
+
+# The SQL generator implementation (default shown)
+#org.opencadc.inventory.db.SQLGenerator=org.opencadc.inventory.db.SQLGenerator
+
+# The schema to use
+org.opencadc.inventory.db.schema=inventory
```
pdowler: The database connection will require schema, username, password, and jdbc url.

Should also be verbose and debug log level option...

The packages to set to those levels (as opposed to default warn) could be listed (single prop with space-separated list of package names) or the packages could be hard coded.... but debug on all of org.opencadc and ca.nrc.cadc is going to be really noisy

cadc-file-validate/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+A file called `cadc-file-validate.properties` should be available from the `/config` directory.  All properties in the file are loaded and set
+as System properties:
+
+```java
+Properties properties = new Properties();
+properties.load("/config/cadc-file-validate.properties");
+System.setProperties(properties);
+```
+
```
pdowler: Please look at some of the other README files and try to follow the same style... they aren't all consistent yet (there's an issue for that) so be careful which you follow.

Also, the storage adapter implementation will require it's own set of properties).... I wonder if the file here should be named java-system.properties so it's clear that it contains config from several different parts.

cadc-file-validate/cadc-file-validate.properties
```diff
@@ -0,0 +1,29 @@
+# Sample configuration file.
```
pdowler: Please don't commit sample config files. Devs can cut & paste from docs and **must edit** their own. Ops will be generating them via some TBD mechanism from something TBD.

Yeah, that's super-vague :-)

cadc-file-validate/README.md
at88mph: This is based off of the Minoc README.  I've updated it a bit.

at88mph: The file contains properties for the storage adapter, the validator, and the database.  We could have a `storage-inventory.properties`, too, but these properties exist together because of the validator.  I'm not sure what the best solution is for the name of the file.

README.md
```diff
@@ -5,8 +5,10 @@ Storage Inventory components
 
 This is the storage inventory data model and architecture documentation.
 
-# cadc-*
+# cadc-file-validate
```
pdowler: Going to come up with a proper name for this tool... TBD

cadc-file-validate/README.md
```diff
@@ -62,12 +62,6 @@ docker build -t opencadc/cadc-file-validate -f Dockerfile .
 docker run -t opencadc/cadc-file-validate:latest
 ```
 
-### With usage
-```
-docker run -t opencadc/cadc-file-validate:latest cadc-file-validate -h
-```
-
-
 ## Running it
```
pdowler: The cadc-java container supports running as nobody; please add/recommend --user nobody:nobody

cadc-file-validate/README.md
```diff
@@ -1,56 +1,56 @@
...
-```
+as System properties.
 
 ### Sample configuration (`cadc-file-validate.properties`)
 
 ```
+#######
+## Common settings
+#######
+org.opencadc.inventory.logging = INFO
```
pdowler: I was thinking this would have to be something like:
```
org.opencadc.{name}.logLevel = INFO
org.opencadc.{name}.logPackages = org.opencadc.inventory ca.nrc.cadc.db ... 
```
in order to be usable to read system props and execute Log4jInit.setLevel(...) calls for an arbitrary set of packages.  Or maybe something like this would allow more flexibility:
```
org.opencadc.{name}.logging.info = org.opencadc.inventory ca.nrc.cadc.db ...
org.opencadc.{name}.logging.debug = org.opencadc.{name} ca.nrc.cadc.net ...
```
The approach has to be usable for the other 3 tools we develop as well and because they bring networking into play there will be more packages to want to fiddle log levels.

at88mph: I see.  Would this be a little more complicated than what is needed?  Would the use case be, "put `cadc-file-validate` in `DEBUG` mode"?  If so, then we can simply say that all `org.opencadc.inventory.*`, `ca.nrc.cadc.net`, `ca.nrc.cadc.util`, `ca.nrc.cadc.db` packages are automatically put into `DEBUG` level.

If that is going to be too noisy, or if we need a finer grained configuration, then maybe we could use what `log4j` does and start with the `logging` keyword to easily find them:

```
logging.org.opencadc.{name} = INFO
logging.org.opencadc.inventory = DEBUG
logging.ca.nrc.cadc.net = TRACE
```

at88mph:
I had not pushed my changes so it was a bit confusing.  Those changes should be reflected now.

There is really only a Template configuration now.  Do we need an Example one below it?

pdowler:
You can have an example below if you want, but I find the really minimal template to be clearer

at88mph:
Great.  Let's leave it as is, then.

Issue #57, opened 2020-02-06
cadc-storage-adapter-fs: must hide temp files until complete


pdowler:
The filesystem adapter writes to a temp file and performs an atomic move to final destination.

However, the temp file is visible within the normal tree during the write so will be picked up by the iterator method (e.g. file-validate process). 

Recommend: divide the configured root directory into two separate sub dirs, e.g. "complete" and "transactions", create temp files in "transactions", and move them to "complete".

hjeeves:
opening issue - tinypm story 2697, task 13024

PR #89, opened 2020-03-10
s2697, ta13024, issue #57: add transaction and content directories for put

hjeeves:
Fixes opencadc/storage-inventory#57

### DIFF
--- a/cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
+++ b/cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
@@ -147,7 +147,7 @@ public class FileSystemStorageAdapter implements StorageAdapter {
     static final String MD5_CHECKSUM_SCHEME = "md5";
     static final int MAX_BUCKET_LENGTH = 5;
     static final int DEFAULT_BUCKET_LENGTH = 2;
-    private static int bucketLength = DEFAULT_BUCKET_LENGTH;
+    private final int bucketLength;
     
     private final FileSystem fs;
     private final Path txnPath;
@@ -191,14 +191,15 @@ public class FileSystemStorageAdapter implements StorageAdapter {
         }
         
         // in uriBucket mode get the bucket depth
+        int bucketLen = 0;
         if (bucketMode.equals(BucketMode.URIBUCKET)) {
             try {
                 String length = pr.getFirstPropertyValue(CONFIG_PROPERTY_BUCKETDEPTH);
                 log.debug("bucketDepth: " + length);
                 if (length != null) {
-                    bucketLength = Integer.parseInt(length);
-                    if (bucketLength < 0 || bucketLength > MAX_BUCKET_LENGTH) {
-                        throw new IllegalStateException("Bucket length of " + bucketLength
+                    bucketLen = Integer.parseInt(length);
+                    if (bucketLen < 0 || bucketLen > MAX_BUCKET_LENGTH) {
+                        throw new IllegalStateException("Bucket length of " + bucketLen
                             + " not in allowed range of 0-" + MAX_BUCKET_LENGTH);
                     }
                 }
@@ -206,6 +207,12 @@ public class FileSystemStorageAdapter implements StorageAdapter {
                 throw new IllegalStateException("failed to load " + CONFIG_PROPERTY_BUCKETMODE
                     + " from " + CONFIG_FILE + ": " + t.getMessage(), t);
             }
+        }
+
+        if (bucketLen == 0) {
+            this.bucketLength = DEFAULT_BUCKET_LENGTH;
+        } else {
+            this.bucketLength = bucketLen;
         }
 
         InventoryUtil.assertNotNull(FileSystemStorageAdapter.class, "rootDirectory", rootVal);
@@ -237,6 +244,7 @@ public class FileSystemStorageAdapter implements StorageAdapter {
         this.contentPath = root.resolve(CONTENT_FOLDER);
         this.txnPath = root.resolve(TXN_FOLDER);
         this.bucketMode = bucketMode;
+        this.bucketLength = DEFAULT_BUCKET_LENGTH;
 
         init(rootDirectory);
     }
--- a/cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
+++ b/cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
@@ -76,7 +76,6 @@ import ca.nrc.cadc.net.ResourceNotFoundException;
 import ca.nrc.cadc.net.TransientException;
 import ca.nrc.cadc.util.HexUtil;
 import ca.nrc.cadc.util.PropertiesReader;
-import ca.nrc.cadc.util.StringUtil;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -91,7 +90,6 @@ import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.LinkOption;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.UserDefinedFileAttributeView;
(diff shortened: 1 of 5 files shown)


cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -317,40 +352,65 @@ public StorageMetadata put(NewArtifact newArtifact, InputStream source)
         InventoryUtil.assertNotNull(FileSystemStorageAdapter.class, "artifact", newArtifact);
         InventoryUtil.assertNotNull(FileSystemStorageAdapter.class, "source", source);
         
-        Path path = null;
-        Path existing = null;
+        Path txnPath = null;
+        Path contentPath = null;
```
pdowler: There should be usable contentPath and transactionPath variables setup in the constructor to use to use... don't recreate it here

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -317,40 +352,65 @@ public StorageMetadata put(NewArtifact newArtifact, InputStream source)
...
+            txnTarget = txnPath.getParent().resolve(UUID.randomUUID().toString());
+            log.debug("resolved txnTarget file: " + txnTarget);
+
+            if (Files.exists(txnTarget)) {
+                // This is an error as the name in the transaction directory should be unique
+                log.debug("file/directory exists");
+                throw new IllegalArgumentException(txnPath + " already exists.");
+            }
+
+            Files.createDirectories(txnTarget.getParent());
```
pdowler: Should never need to createDirectories for txnTarget

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -317,40 +352,65 @@ public StorageMetadata put(NewArtifact newArtifact, InputStream source)
...
+
+        } catch (InvalidPathException e) {
+            throw new IllegalArgumentException("Illegal path: " + txnPath, e);
+        } catch (IOException e) {
+            throw new IllegalStateException("Failed to create transaction file: " + txnTarget, e);
+        }
+
```
pdowler: The try block below (createStorageLocationPath) should be delayed until the ioLoop finishes... closer to the move so less chance of creating junk in contentPath (probably right after the setFileAttribute and right before the atomic move).

Check the docs for Files.createDirectories but iirc it is idempotent (like mkdir -p); if that's the case it would be safe and faster to just blindly call it (it will traverse the path once instead of twice).

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -396,7 +456,12 @@ public StorageMetadata put(NewArtifact newArtifact, InputStream source)
             }
 
             // Set contentChecksum file attribute
-            setFileAttribute(path, CHECKSUM_ATTRIBUTE_NAME, checksum.toString());
+            setFileAttribute(txnTarget, CHECKSUM_ATTRIBUTE_NAME, checksum.toString());
+
+            // to atomic copy into content directory
+            Path newCopy = Files.move(txnTarget, contentPath, StandardCopyOption.ATOMIC_MOVE);
+            log.debug("moved file to : " + contentPath);
```
pdowler: You also want REPLACE_EXISTING in the options to move. The previous code overwrote the existing file content and only used move to "rollback"... now this code relies on move to replace.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
```diff
@@ -490,6 +549,8 @@ public void delete(StorageLocation storageLocation)
     public SortedSet<StorageMetadata> list(String storageBucket)
         throws StorageEngageException, TransientException {
         StringBuilder path = new StringBuilder();
+        path.append(CONTENT_FOLDER);
```
pdowler: Use contentPath (after rename) directly here as the parent. Traversal through subdirectories should be always be done via 
```Path p = contentPath.resolve(sbpath);```

File: cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java (hunk 4 above)
Comment (hjeeves) [reply cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java]: ok

File: cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java (hunk 2 above)
Comment (hjeeves) [reply cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java]: ok

File: cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java (hunk 1 above)
Comment (hjeeves) [reply cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java]: ok

File: cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java [hunk 6]
```diff
@@ -141,16 +143,20 @@
...
     static final int DEFAULT_BUCKET_LENGTH = 2;
     
     private FileSystem fs;
     private Path root;
+    private Path contentFolder;
+    private Path transactionFolder;
     private BucketMode bucketMode;
     private int bucketLength = DEFAULT_BUCKET_LENGTH;
-    
+
```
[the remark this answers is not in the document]
hjeeves: ok

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
hjeeves: believe I've got this... the storage bucket part of the path is still built in a string builder (particularly the URIBUCKET case,) - but then that string is added via .resolve to the contentPath.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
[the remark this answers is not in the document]
pdowler: bucketMode can also be final. You can (should) move the assignment from init() to ctor(s).

cadc-storage-adapter-fs/src/test/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapterTest.java
```diff
@@ -111,7 +112,7 @@
     private static final byte[] data = dataString.getBytes();
 
     static {
-        Log4jInit.setLevel("org.opencadc.inventory", Level.INFO);
+        Log4jInit.setLevel("org.opencadc.inventory", Level.DEBUG);
```
pdowler: Prefer to have these at INFO in master; devs turn it up to DEBUG if they need more details than INFO provides

hjeeves: done

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/FileSystemStorageAdapter.java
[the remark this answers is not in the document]
hjeeves: done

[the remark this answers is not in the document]
pdowler: just noticed that bucketLength is static instead of final. While you are at it - it should be mandatory with no default value for the mode that uses it (enforcing the MAX_BUCKET_LENGTH is fine -- 5 means 1e6 directories at level 5)

PR #97, opened 2020-03-19
cadc-inventory-db ArtifactDAO iterator fixes 

pdowler:


cadc-inventory-db/src/main/java/org/opencadc/inventory/db/SQLGenerator.java
```diff
@@ -467,31 +473,46 @@ public void setPrefix(String prefix) {
...
+            log.debug("sql: " + sql);
             
             try {
                 Connection con = ds.getConnection();
+                log.debug("ArtifactIterator: setAutoCommit(false)");
+                con.setAutoCommit(false);
+                // default options: ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY
                 PreparedStatement ps = con.prepareStatement(sql);
-                ps.setFetchSize(1000);
+                ps.setFetchSize(20);
```
andamian: Why only 20?

cadc-inventory-db/src/main/resources/inventory.vacuum-analyze.sql
```diff
@@ -0,0 +1,6 @@
+-- periodic DB 
+VACUUM VERBOSE ANALYZE <schema>.Artifact;
```
andamian: What is this for?

cadc-inventory-db/src/main/java/org/opencadc/inventory/db/SQLGenerator.java
pdowler: was for testing... reverted to 1000.

PR #103, opened 2020-03-24
s2575, ta 13130: parse critwall config file

hjeeves:


critwall/src/main/java/org/opencadc/critwall/Main.java
```diff
@@ -67,51 +67,198 @@
...
             Log4jInit.setLevel("org.opencadc.reg", cfg);
-            
-            // parse config file and populate/assign these
+            Log4jInit.setLevel("org.opencadc.critwall", cfg);
+
+            log.debug("log level: " + cfg);
+            log.debug("properties: " + props.toString());
+
+            // parse config file
+            String errMsg = "";
```
pdowler: Use a StringBuilder and append when there are more than a couple of strings ops to do.

critwall/src/main/java/org/opencadc/critwall/Main.java
```diff
@@ -67,51 +67,198 @@
...
             System.exit(-1);
+
         }
+        log.debug("finished critwall run.");
     }
     
     private Main() { 
     }
+
+    static final String getSingleProperty(MultiValuedProperties props, String key) {
```
pdowler: getSingleProperty no longer needed - use MultiValuedProperties.getFirstPropertyValue directly

critwall/src/main/java/org/opencadc/critwall/Main.java
hjeeves: ok

critwall/src/main/java/org/opencadc/critwall/Main.java
hjeeves: done

PR #104, opened 2020-03-26
S2655 - update minor and raven to use PermissionsClient with baldur

jburke-cadc:


raven/build.gradle
```diff
@@ -41,9 +41,10 @@ dependencies {
     compile 'org.opencadc:cadc-inventory:[0.1,)'
     compile 'org.opencadc:cadc-inventory-db:[0.1,)'
     compile 'org.opencadc:cadc-storage-permissions:[0.1,)'
+    compile 'org.opencadc:cadc-storage-permissions-client:[0.1,)'
     compile 'org.opencadc:cadc-vos:[1.1.8,)'
-    
-    runtime 'org.opencadc:cadc-access-control:[1.1.1,2.0)'
+    compile 'org.opencadc:cadc-access-control:[1.1.1,2.0)'
```
pdowler: The compile dependency should be cadc-gms and use that API in the code. For now, we do need to include a runtime dependency on cadc-access-control for the implementation.

Same for the minoc/build.gradle as well

raven/src/main/java/org/opencadc/raven/PostAction.java
```diff
@@ -236,43 +298,37 @@ public void doAction() throws Exception {
...
-            try {
-                String sqlGenClass = sqlGenList.get(0);
-                cls = Class.forName(sqlGenClass);
-            } catch (ClassNotFoundException e) {
-                throw new IllegalStateException("could not load SQLGenerator class: " + e.getMessage(), e);
-            }
-        } else {
-            // use the default SQL generator
-            cls = SQLGenerator.class;
+    private void checkReadPermission(URI artifactURI) throws AccessControlException, ResourceNotFoundException,
```
pdowler: Same comment about ResourceNotFoundException from minoc/ArtifactAction applies here...

minoc/src/main/java/org/opencadc/minoc/ArtifactAction.java
```diff
@@ -320,16 +314,13 @@ private void initStorageSite(URI resourceID) {
...
+            throw new IllegalStateException("BUG: found " + curlist.size() + " StorageSite entries");
         }
         log.info("initStorageSite: " + resourceID + " " + name);
         SELF_RESOURCE_ID = resourceID;
     }
     
-    public void checkReadPermission()
-        throws AccessControlException, ResourceNotFoundException,
-               TransientException {
+    public void checkReadPermission() throws AccessControlException, ResourceNotFoundException, TransientException {
```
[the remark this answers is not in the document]
jburke-cadc: Looking at baldur, and what I just did in ams, if an artifactURI is not found, an empty grant with no permissions is returned (anon = false, empty groups list). The client should never throw a ResourceNotFoundException. An empty grant will result in an AccessControlException, read|write permission denied. The client never exposes that the resource was not found.

raven/build.gradle
jburke-cadc: Updated.

cadc-storage-permissions-client/src/main/java/org/opencadc/inventory/permissions/PermissionsClient.java
```diff
@@ -114,11 +114,10 @@ public PermissionsClient(URI serviceID) {
      *
      * @param artifactURI Identifies the artifact for which to retrieve grant information.
      * @return The read grant information.
-     * 
-     * @throws ResourceNotFoundException If the file could not be found.
+     *
```
pdowler, line 117: The possible return of null needs to be documented in the @return since it is meaningful  and must be interpretted (and in getWriteGrant)

minoc/src/main/java/org/opencadc/minoc/ArtifactAction.java
```diff
@@ -265,7 +265,13 @@ public void checkReadPermission()
     
     public void checkWritePermission()
         throws AccessControlException, ResourceNotFoundException, TransientException {
-        
+
```
pdowler, line 268: In master the check for anon comes before the hack to allow unrestricted write during development. That was on purpose to require authentication but skip authorization. Why has that changed here?

raven/src/main/webapp/META-INF/context.xml
```diff
@@ -9,9 +9,9 @@
         factory="org.apache.tomcat.jdbc.pool.DataSourceFactory" closeMethod="close"
         minEvictableIdleTimeMillis="60000" timeBetweenEvictionRunsMillis="30000"
         maxWait="20000"
-        initialSize="0" minIdle="0" maxIdle="${raven.invuser.maxActive}" maxActive="${raven.invuser.maxActive}"
-        username="${raven.invuser.username}" password="${raven.invuser.password}"
-        driverClassName="org.postgresql.Driver" url="${raven.invuser.url}"
+        initialSize="0" minIdle="0" maxIdle="${raven.invadm.maxActive}" maxActive="${raven.invadm.maxActive}"
```
pdowler, line 12: This is wrong. raven only queries the database and the intent is that it uses the account with the lowest priv (tapuser)... hence the prop name invuser

raven/src/main/java/org/opencadc/raven/InitDatabaseAction.java
```diff
@@ -0,0 +1,247 @@
+/*
```
pdowler, line 1: Why was this whole file copied from minoc? raven is not responsible for creating/updating the database and it's obvious that initStorageSite() is completely wrong at a global inventory site. 

I know we agreed to copy the config handling but that's only a subset of this class and it is in a separate class in minoc because that service does need to init on startup. An InitAction might be a nice way to achieve "read once" config here, but this class does not do that and is not even executed by adding it to web.xml ... so as it is we **might only** have a really confusing way to add a static method for reading config.

cadc-storage-permissions-client/src/main/java/org/opencadc/inventory/permissions/PermissionsClient.java
jburke-cadc: Added documentation that a null could be returned.

minoc/src/main/java/org/opencadc/minoc/ArtifactAction.java
jburke-cadc: I think I copied the hack in front of the auth check for a quick test, then forgot to remove it. Removed.

raven/src/main/webapp/META-INF/context.xml
jburke-cadc: The raven readme used invadm, and pgdev had an invadm user, but not an invuser. I thought invuser was incorrect. Should've asked. Updated the readme and the context.xml to use invuser.

raven/src/main/java/org/opencadc/raven/InitDatabaseAction.java
jburke-cadc: I misunderstood then. I thought the idea was copy the class, then later move common setup/config code to a util library. Removed the class.

PR #106, opened 2020-03-31
s2575, ta13144: BucketSelector implementation

hjeeves:


critwall/src/main/java/org/opencadc/critwall/BucketSelector.java
```diff
@@ -67,18 +67,71 @@
...
+                log.debug("min toShort failed" + e);
+                throw new IllegalArgumentException("invalid format: " + selectors, e);
+            }
+        }
+
+        if (min < 0 || max < min || max > 15) {
+            throw new IllegalArgumentException("invalid bucket selector (min,max): " + min + "," + max);
+        }
+    }
+
```
pdowler, line 120: I assume min and max is the API that FileSync will use to get values for calling ArtifactDAO.unstoredIterator(String), so FileSync needs string bucket prefixes.

Why are the int(s) exposed? What are the int(s)? 

Say the buckets value is "a - d"... that means FileSync has to call unstoredIterator(bp) with values bp = a, b, c, d... so some code is going to have to create a sequence of strings with those 4 bucket prefixes. I would put that sequence generation in BucketSelector.

sequence is either a List<String> or an Iterator<String>... since it is code-generated, there isn't much point in creating all the values up front (list)... so Iterator it is. These four method could be replaced by:
```
public Iterator<String> getBuckets()
```
and then you could implement a little of FileSync.run() to see if the approach makes sense. Something like:
```
Iterator<String> buckets = bs.getBuckets();
while (buckets.hasNext()) {
   String b = buckets.next();
   log.info("processing bucket: " + b);
}
throw new UnsupportedOperationException("TODO");
```
(I put the throws after the loop so multiple log messages can be seen when the container is run)

It's always better to get the connectivity between things worked out correctly early on and write at least some of the code on both sides (the API and the use of the API) at the same time.

critwall/src/main/java/org/opencadc/critwall/BucketSelector.java
```diff
@@ -67,18 +67,71 @@
...
+    private final String hexMax;
+    private final String hexBuffer = "000";
 
-    BucketSelector(String selectorPrefix) {
-        this.bucketSelector = selectorPrefix;
-        throw new UnsupportedOperationException("TODO");
+    public BucketSelector(String selectors) {
+        InventoryUtil.assertNotNull(BucketSelector.class, "selectors", selectors);
+        this.bucketSelectors = selectors;
+
```
pdowler, line 91: In the ctor you will want to split the input, validate the 1 or 2 values, and save. For now, it would be OK to force both values to be the same length and I'm fine with forcing them to be length 1 to start. 

Validate just means that the chars in the hex char set [0-9,a-f]... with odd number of chars you can't use HexUtil without padding. If it was me I'd brush up on the UTF-8 charset order and see if I could use 'char' arithmetic. It is probably better to think about them as strings or even char[ ]

note: 
0-f is 16 buckets, might just be `for (char c = '0', c <= 'f'; c++)`
0 os one bucket (1/16th of the space)
00 is one bucket 00 (1/256 of the space) and this not equivalent to 0
0-15 is 7 buckets: 0, 10, 11, 12, 13, 14, 15
16-f is the the remaining 12+14 = 26 buckets: 16, 17, ... 1f, 2, 3, ... f
 (these last two subdivide bucket 1 unevenly as 10-16 and 17-1f)

hjeeves: a quick test of char math for Java produces 0-9 :;<=>?@ABC...abcd. doesn't appear hex-friendly. :(

critwall/src/test/java/org/opencadc/critwall/BucketSelectorTest.java
```diff
@@ -0,0 +1,174 @@
...
+public class BucketSelectorTest {
+
+    private static final Logger log = Logger.getLogger(BucketSelectorTest.class);
+
+    static {
+        Log4jInit.setLevel("org.opencadc.critwall", Level.DEBUG);
+    }
+
+    public BucketSelectorTest() { }
+
```
pdowler, line 90: For all of these tests, I would add more tests to verify whitespace handling: space before, space between, spaces after. Can be multiple checks in sequence in testValidRange and testValidSingleSelector.

There seems to be code to handle config like " - 2" and "2 - " but no tests with such values. testRangeSyntax?

In general, in a test for an invalid input, make nothing but the thing that's wrong and being tested is there to obscure the intent. For example, line 133 has "5- jj" which has 3 things that catch the eye:
- asymmetric spacing (harmless but distracting)
- second value is too long (cause fail)
- second value has invalid characters (cause fail)
So - this has two things wrong in a single test so I can't tell which code path was tested: should be two separate tests with one thing wrong each: testInvalidCharacters? testInvalidBucketLength?

critwall/src/main/java/org/opencadc/critwall/BucketSelector.java
```diff
@@ -67,18 +67,83 @@
...
 
-/**
- *
- * @author pdowler
- */
 public class BucketSelector {
-    private final Logger log = Logger.getLogger(BucketSelector.class);
-    public final String bucketSelector;
+    private static final Logger log = Logger.getLogger(BucketSelector.class);
+    // range as entered in .properties file
```
pdowler, line 79: First thing: minimise member variables
-  in this impl you only need the bucketList once the ctor is done, so that should be final; you do not need to keep bucketSelectors, rangeMin,  and rangeMax (*unless*: classes with state should override toString to assist debugging in FileSync and that should output "BucketSelector[" + rangeMin + "," + rangeMax +"]" so that could be a reason to keep those two member variables)
- maybe checkstyle is not enforcing this but static declarations are supposed to come before non-static and it helps comprehension to put them first

Just to make the limitations if this impl clear, I would declare a constant 
`private static final int MAX_PREFIX_LENGTH = 1;` 
to use in the ctor validation. 

I am also considering that the HEXVALUES constant should be declared in Artifact rather than here. If `public static final String Artifact.URI_BUCKET_CHARS` had been defined there, with a name like that, would it has made this whole thing easier to grok?

critwall/src/main/java/org/opencadc/critwall/BucketSelector.java
```diff
@@ -67,18 +67,83 @@
...
+
+    public Iterator<String> getBucketIterator() {
+        return bucketList.iterator();
+    }
+
+    public BucketSelector(String selectors) {
+        InventoryUtil.assertNotNull(BucketSelector.class, "selectors", selectors);
+        this.bucketSelectors = selectors;
+
+        String[] minMax = this.bucketSelectors.split("-");
```
pdowler, line 96: line 97: StringBuffer is thread safe and performs poorly; always use StringBuilder unless you need thread safety (same API)

line 106 and 120: why not minMax[i].trim()? don't apply toLowerCase - values in config should be lower case or fail the subsequent validation; the storage dm specifies what the values look like and upper and lower case are *not* equivalent

line 108 and 122: these check for a token that was trimmed to 0-length? but the lines 111 and 125 look for those empty strings in the alphabet anyway... the error message will be "invalid hex value: \n" (confusing)

Layerrs of nested if/else are really hard to follow. We used to have a code-style check that counted the number of different paths through a method (~complexity) and nested if/else is remarkably good at creating excess complexity. More straightforward would be to:
```
rangeMin = minMax[0].trim();
rangeMax = rangeMin;
if (minMax.length == 2) {
   rangeMax = minMax[1].trim();
}
// validate length of tokens

// validate characters in tokens

// validate order

// construct set
```

where each section completes or throws an exception (which also reduces nested if/else - for example, the else on line 104 can be removed because minMax.length has to be 1 or 2 for all subsequent code)

critwall/src/main/java/org/opencadc/critwall/BucketSelector.java
hjeeves: Oh yeah.. if HEXVALUES had been in Artifact, it could have steered understanding. If it's likely it will be needed by other apps in the storage-inventory corral, having it central like that would be helpful.

hjeeves: Based on what you have there, I'll keep rangeMin & rangeMax.

critwall/src/main/java/org/opencadc/critwall/BucketSelector.java
```diff
@@ -67,18 +67,88 @@
...
+            errMsg.append("empty max range value: " + rangeMax + "\n");
+        }
+
+        if (rangeMin.length() > MAX_PREFIX_LENGTH) {
+            errMsg.append("min range value greater than maxlen (" + MAX_PREFIX_LENGTH + "): " + rangeMin + "\n");
+        }
+        if (rangeMax.length() > MAX_PREFIX_LENGTH) {
+            errMsg.append("max range value greater than maxlen (" + MAX_PREFIX_LENGTH + "): " + rangeMax + "\n");
+        }
+
```
pdowler, line 123: - last review said "where each section completes or throws an exception"; otherwise, down on line 127 you are still looking in the Artifact.URI_BUCKET_CHARS string with invalid input

- the 4 `if` statements (line 110 to 122) can be reduced to

```
if (rangeMin.length() != MAX_PREFIX_LENGTH || rangeMax.length() != MAX_PREFIX_LENGTH) {
    throw new IllegalArgumentException("invalid config: ....");
}
```
If you want to make it more clear that in principle that is a max and not the only length, it could be
```
if ((1 <= rangeMin.length() && rangeMin.length() <= MAX_PREFIX_LENGTH)
    && (1 <= rangeMax.length() && rangeMax.length() <= MAX_PREFIX_LENGTH) {
    log.debug("acceptable length: " + rangeMin + " - " + rangeMax);
} else {
    throw new IllegalArgumentException("invalid config: ....");
}
```

critwall/src/main/java/org/opencadc/critwall/BucketSelector.java
```diff
@@ -67,18 +67,88 @@
...
+        // 0-f is acceptable range
+        int min;
+        int max;
+        min = Artifact.HEXVALUES.indexOf(rangeMin);
```
pdowler, line 127: - lines 127 and 131: in the case where rangeMin.length() == 1, Artifact.URI_BUCKET_CHARS.indexOf(String) is OK, but this check for the index of rangeMin isn't generally correct; in general, each character in rangeMin needs to be looked up in the Artifact.URI_BUCKET_CHARS individually to validate rangeMin (and rangeMax)... you can decide if a comment explaining the short cut or a char-by-char validation is needed

hjeeves: The algorithm for populating bucketList would need to change as well if MAX_PREFIX_LENGTH > 1.  Because of that, I'm opting for adding a comment for both this and the section where bucketList is populated.

critwall/src/main/java/org/opencadc/critwall/BucketSelector.java
hjeeves: - compressed those lines, added more general check
- exceptions thrown rather than compiling into StringBuilder

critwall/src/main/java/org/opencadc/critwall/BucketSelector.java
```diff
@@ -67,18 +67,88 @@
...
- */
 public class BucketSelector {
-    private final Logger log = Logger.getLogger(BucketSelector.class);
-    public final String bucketSelector;
+    private static final Logger log = Logger.getLogger(BucketSelector.class);
+
+    private static final int MAX_PREFIX_LENGTH = 1;
+    private final String rangeMin;
+    private final String rangeMax;
+    private final TreeSet<String> bucketList = new TreeSet<String>();
```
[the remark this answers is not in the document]
hjeeves: - added toString() function and added it into unit tests.

cadc-inventory/src/main/java/org/opencadc/inventory/Artifact.java
```diff
@@ -84,6 +84,7 @@
  */
 public class Artifact extends Entity {
     private static final Logger log = Logger.getLogger(Artifact.class);
+    public static final String HEXVALUES = "0123456789abcdef";
```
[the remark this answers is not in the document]
hjeeves: - renamed in cadc-inventory

PR #112, opened 2020-04-09
implemented scalable OpaqueFileSystemStorageAdapter

pdowler:


cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/OpaqueIterator.java
```diff
@@ -0,0 +1,223 @@
...
+    private final Path contentPath;
+    private Path subPath;
+    
+    private StorageMetadata nextItem = null;
+    private final LinkedList<StackItem> stack = new LinkedList<>();
+    
+    public OpaqueIterator(Path contentPath, String bucketPrefix) throws StorageEngageException {
+        this.contentPath = contentPath;
+        StringBuilder sb = new StringBuilder();
+        if (bucketPrefix != null) {
```
at88mph, line 105: Is the bucket optional for those iterator() methods that don't specify a bucket?  Does this imply that Tantar can make a full run across all buckets?

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/OpaqueIterator.java
```diff
@@ -0,0 +1,223 @@
...
+ * @author pdowler
+ */
+class OpaqueIterator implements Iterator<StorageMetadata> {
+    private static final Logger log = Logger.getLogger(OpaqueIterator.class);
+
+    private final Path contentPath;
+    private Path subPath;
+    
+    private StorageMetadata nextItem = null;
+    private final LinkedList<StackItem> stack = new LinkedList<>();
```
at88mph, line 100: Picky, but can this be renamed to what it does?  It _appears_ to maintain the tree traversal, but a little confusing.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/OpaqueIterator.java
pdowler: Yes, null bucketPrefix matches everything. I will make that explicit in the API javadocs.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/OpaqueIterator.java
pdowler: It is a stack (FIFO) to support depth-first traversal, but the actual Stack class is not recommended for use because it is based on the old Vector class that makes heavy use of synchronized blocks. I'll try to come up with a better var name so it is more easily understood.

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/OpaqueIterator.java
pdowler: Javadoc already says bucketPrefix may be null, partial, or complete storageBucket

PR #113, opened 2020-04-15
Reporter updates with a stub for a RecoverFromStoragePolicy.

at88mph:


tantar/src/main/java/org/opencadc/tantar/policy/RecoverFromStoragePolicy.java
```diff
@@ -0,0 +1,108 @@
...
+     * Use the logic of this Policy to correct a conflict caused by the two given items.  One of the arguments can
+     * be null, but not both.
+     *
+     * @param artifact        The Artifact to use in deciding.
+     * @param storageMetadata The StorageMetadata to use in deciding.
+     * @throws Exception For any unknown error that should be passed up.
+     */
+    @Override
+    public void resolve(final Artifact artifact, final StorageMetadata storageMetadata) throws Exception {
+        if (artifact != null) {
```
pdowler, line 100: This simple implementation has to consider the scenario where both artifact and storageMetadata exist (eg it was run, killed/died, and restarted and is redoing some previously recovered entries). So it does need to handle both args not null as a potential replace or update.

at88mph: Thanks.  If both args are not null, it will be assumed that the policy wasn't switched or anything, it just wasn't completed.

at88mph:
Both the `createArtifact` and `replaceArtifact` methods in the `BucketValidator` class already use all seven of the fields from the `StorageMetadata` class to create the Artifact.  Should those methods reset the fields that don't exist in the `StorageMetadata`?

tantar/src/main/java/org/opencadc/tantar/policy/RecoverFromStoragePolicy.java
pdowler: Full replace is wrong if it is just fixing some mutable metadata. It will ultimately cause other sites to delete files and re-download them when they didn't change.

pdowler:
In RecoverFromStorage they should set everything mutable to be the value from storage, including replacing non-nmull with null (eg contentEncoding). It is intended to be equivalent to "wipe and start over". By updating (rather than replacing) the artifact, the immutable fields (contentChecksum and contentLength specifically) mena that other sites won't have to download the file again.

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
```diff
@@ -543,6 +544,47 @@ public void replaceArtifact(final Artifact artifact, final StorageMetadata stora
...
+                LOGGER.debug("Start transaction.");
+                transactionManager.startTransaction();
+
+                // By reusing the Artifact instance's ID we can have the original Artifact but wipe out the mutable
+                // fields below.
+                artifact.contentEncoding = storageMetadata.contentEncoding;
+                artifact.contentType = storageMetadata.contentType;
+                artifact.storageLocation = storageMetadata.getStorageLocation();
+
+                artifactDAO.put(artifact, true);
```
pdowler, line 570: In this case you can do a plain put(artifact) since you know the storageLocation is equal/correct and are only updating artifact metadata... that means the DAO computes checksum and only updates lastModified if the checksum changes. I think that is correct here. 

In fact, it further illustrates that **this** policy doesn't really need to check anything in this case: it could assign metadata values and do the put for every artifact and not have to replicate checking for differences...

at88mph: I modified the `put(Artifact)` call.

PR #114, opened 2020-04-17
s2575, ta13128: FileSyncJob first code review

hjeeves:


critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+            // check that artifact in db has storage location
+            Assert.assertNotNull(storedArtifact.storageLocation);
+
+            StorageLocation expected = new StorageLocation(new URI("ad:IRIS/I212B2H0.fits"));
+            expected.storageBucket = "ad:IRIS";
+            Assert.assertEquals(expected, storedArtifact.storageLocation);
+
+            // check for file on disk
+            ByteArrayOutputStream dest = new ByteArrayOutputStream();
```
pdowler, line 203: If line 177 is correct, this is ~1MiB of ram. Certainly modest and probably unnoticeable... except if grabage collection doesn't happen right away or other things we don't know/care about do happen.
At least make the intent to discard clear by calling dest.reset() after the get.

If you didn't want to worry about the file size, `new FileOutputStream("/dev/null")` would work and the intent would be clear :-)

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
```diff
@@ -0,0 +1,213 @@
...
+            fsj.run();
+
+            // assert that job succeeded
+            // Get the artifact by URI again
+            Artifact storedArtifact = dao.get(artifactID);
+
+            // check that artifact in db has storage location
+            Assert.assertNotNull(storedArtifact.storageLocation);
+
+            StorageLocation expected = new StorageLocation(new URI("ad:IRIS/I212B2H0.fits"));
```
pdowler, line 198: I would not assert anything about the value of the storageLocation and bucket; those are supposed to be opaque and any change in the adapter will break this test. As long as it is not null and the get works, that's all that matters. 

The verification below should therefore be `sa.get(storedArtifact.storageLocation, dest)`

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -0,0 +1,212 @@
...
+            FileContent content = new FileContent(out.toByteArray(), "text/xml");
+            log.debug("xml file content to be posted: " + transfer);
+
+            log.debug("artifact path: " + artifactID.getPath());
+            HttpPost post = new HttpPost(certURL, content, true);
+            post.prepare();
+            log.debug("post prepare done");
+
+            // TODO: error handling is likely very different than this...
+            if (post.getThrowable() != null && post.getThrowable() instanceof FileNotFoundException) {
```
pdowler, line 149: post.getThrowable() is null if you call prepare() because prepare() throws exceptions instead of capturing them; this can be removed.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -0,0 +1,212 @@
...
+                throw (FileNotFoundException) post.getThrowable();
+            }
+
+            TransferReader reader = new TransferReader();
+            Transfer t = reader.read(post.getInputStream(), null);
+            List<String> urlList = t.getAllEndpoints();
+
```
pdowler, line 156: Ouch - getAllEndpoints() shows how dated and sloppy the cadc-vos lib still is :-)
I think you could move all the code above up to this point into a method, maybe
`private List<URL> getDownloadURLs(...)`
The negotiation part cannot really be retried (except for some transient failures that HttpPost already retries) so if it fails the job needs to give up and return.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -0,0 +1,212 @@
...
+            log.debug("endpoints returned: " + urlList);
+            // get the URL list from t, then call HttpGet to get the item
+
+            int retryCount = 0;
+            boolean found = false;
+
+            for (int i = 0; i < urlList.size(); i++) {
+                URL u = new URL(urlList.get(i));
+                log.debug("trying this url: " + u);
+
```
pdowler, line 166: If this `while (retryCount` loop was outside the `for (URL u : urlList)` loop, it could be a good strategy for cases where some sites/servers are loaded to the point of failure and others are not... so not discounting it if those loops were swapped. As is, it tries a URL 2x with no delay and then moves on.

However, HttpGet already has a retry mechanism built in (it happens in `get.prepare()`) with a gradually increasing delay (unless the server says how long to delay) -- and delay between retries is absolutely necessary because otherwise the retries just make the situation worse instead of better. By default it only retries for transient exceptions, which is fine for now (it can be changed). The number of retries can be set (check code for default but I think it is in the ~3). 

The second thing about a retry loop is you have to identify when a retry might give a different result and when it will not. Here, I think that means a try/catch around the sequence from line 169 to 176 where you specifically deal with each exception that HttpGet.prepare() and StorageAdapter.put() might throw: 
* if the failure from prepare() is fatal (eg. ResourceNotFoundException ) then remove that URL from  the list
* if the failure from put() is fatal (eg WriteException or StorageEngageException?) then abort the whole job
Feel free to use multicatch style to handle several different exceptions in the same way.

This is the hard part because it is very hard to simulate (make a test that provokes) most of the failures so it comes down to thorough analysis (code review) to have good error handling.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -0,0 +1,212 @@
...
+                while (retryCount < 2 && found == false) {
+                    ByteArrayOutputStream dest = new ByteArrayOutputStream();
+                    HttpGet get = new HttpGet(u, dest);
+                    get.prepare();
+                    Artifact curArtifact = artifactDAO.get(artifactID);
+
```
pdowler, line 172: The above artifactDAO.get is too early (before the data transfer); it should happen farther down: right before curArtifact is used. That's not technically sufficient to prevent all race conditions, but it will make it easier to add the right locking later on.

critwall/src/test/resources/cadc-storage-adapter-fs.properties
```diff
@@ -0,0 +1,11 @@
+# The root of the file system (as seen by the container)
```
pdowler, line 1: I don't think this file is needed since FileSyncJobTest constructs the adapter via the utility ctor... if it was used it would belong in intTest/resources.

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
hjeeves: went for /dev/null - done

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
hjeeves: done

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: done

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: Put this code into a function, try/catch around function call to quit run early if this fails.
Decided to prune malformed URLs from the list at this point: as long as one is good, something can be tried.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: This led to a larger rework. The 'sync' part itself (HttpGet & storage adapter put) are in a function with a loop over the URL list. The outer while loop calls this function, and stops if the storage metadata is returned successfully, or its max number of retries is reached (in retrospect, maybe that number should be tuneable in the properties file?)

Split out the exceptions by retry or fail, and also by get or put to some degree. Maybe the latter isn't necessary as the log messages will have details. (?)

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: This has been moved, in the rework done in response to your previous comment.

critwall/src/test/resources/cadc-storage-adapter-fs.properties
hjeeves: file is gone. was a remnant from having the FileSyncJobTest in the 'test' folder instead of 'intTest'

critwall/build.gradle
```diff
@@ -21,12 +21,13 @@ dependencies {
     compile 'org.opencadc:cadc-storage-adapter:[0.3,1.0)'
     compile 'org.opencadc:cadc-util:[1.3.4,2.0)'
     compile 'org.opencadc:cadc-inventory:[0.7,2.0)'
-    compile 'org.opencadc:cadc-inventory-db:[0.6,1.0)'
+    compile 'org.opencadc:cadc-inventory-db:[0.7.4,1.0)'
     compile 'org.opencadc:cadc-registry:[1.5,2.0)'
-    compile 'org.opencadc:cadc-storage-adapter-ad:[0.1,)'
+    compile 'org.opencadc:cadc-storage-adapter-fs:[0.6,)'
+    compile 'org.opencadc:cadc-vos:[1.1.8,)'
+    compile 'org.opencadc:cadc-vosi:[1.0,)'
```
pdowler, line 28: are you sure cadc-vosi is needed here? It's generally a (poorly named) server-side lib

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -0,0 +1,256 @@
...
+                log.info("Malformed URL returned from transfer negotiation: " + urlStrList.get(i) + " skipping... ");
+            }
+        }
+
+        return urlList;
+    }
+
+    private StorageMetadata syncArtifact(Iterator<URL> urlIterator) {
+        StorageMetadata storageMeta = null;
+
```
pdowler, line 215: This looks pretty good now and is easy to grok. I would still improve it further with:

line ~228 : immediately return the StorageMetadata object rather than `break`
line ~230 : in the first catch, use multi-catch for all the exceptions where you would just continue trying (TransientException, ReadException,  IOException for now), don't need `continue`
line ~235 : you should call urlIterator.remove() here so that URL is gone if we ever add any kind of pause/resume or longer retry delay
line ~239 : I would just have this method let the fatal errors propagate up (StorageEngageException, WriteException, InterruptedException)

With that, you won't need the return at the end and can never return null.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -0,0 +1,256 @@
...
+        } catch (Exception e) {
+            // fail - nothing can be done without the list, negotiation cannot be retried
+            log.error("transfer negotiation failed.");
+            return;
+        }
+        log.debug("endpoints returned: " + urlList);
+
+        int retryCount = 0;
+        Iterator<URL> urlIterator = urlList.iterator();
+
```
pdowler, line 141: with the changes in syncArtifact, this `while` loop needs to be wrapped in a try/catch for then fatal errors. The catch can be more or less like the one for failed negoitiaton (log and return -- and I do like the return in there so it is explicit that nothing more can be done even when it is near or at theb end of run()). 

Since the HttpGet.prepare() inside syncArtifact does some retrying already (mainly 503s from server), the while loop here is mainly compensating for failed in transit (ReadException during the put)... with any retry loop you need a delay (Thread.sleep(milliseconds) which should be down by the `retryCount++`... a compact way to do it with increasing delay is declare a constant like:
```
static long[] RETRY_DELAY = new long[] { 2000L, 4000L };
```
then use `while (retryCount < RETRY_DELAY.length)` and `Thread.sleep(RETRY_DELAY[retryCount++]);` and increasing the number of retries means adding delays to the array.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: IOException will catch the WriteException (which would turn it into a transient instead of a failure situation), if I'm reading the error my IDE reports correctly?

Returning null is the only thing that would trigger the outer while loop to retry the entire list. If no storageMeta is returned, then none of the URLs resulted in a successful get. The catastrophic fails already quit, but if there are transient errors for each url, there has to be a case where a null can fall through and trigger a retry on the whole list. (Or am I misunderstanding the retry idea?)

critwall/build.gradle
hjeeves: (as discussed in slack) gone.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: changes made, but the WriteException is not handled symmetrically with the rest of the exceptions because of the IOException swallowing it unless it's trapped explicitly.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: done

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -0,0 +1,258 @@
...
+                ByteArrayOutputStream dest = new ByteArrayOutputStream();
+                HttpGet get = new HttpGet(u, dest);
+                get.prepare();
+
+                NewArtifact a = new NewArtifact(artifactID);
+                storageMeta = storageAdapter.put(a, get.getInputStream());
+                log.debug("storage meta returned: " + storageMeta.getStorageLocation());
+                return storageMeta;
+
+            } catch (WriteException wre) {
```
pdowler, line 239: I think we have to consider WriteException fatal so it needs to be thrown (retry isn't likely to help in the short term). You can just throw it from the catch and subsequent catches don't interfere.

So returning null means that all URLs were tried without success but without a fatal error. I guess that works but please add a comment for the method like:
```
// success: return StorageMetadata
// fail: return null if all URLs failed | throw if failure is fatal (internal)

File: critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java [hunk 12]
```diff
@@ -0,0 +1,229 @@
...
+            DBUtil.createJNDIDataSource("jdbc/FileSyncJobTest", cc);
+
+            Map<String,Object> config = new TreeMap<String,Object>();
+            config.put(SQLGenerator.class.getName(), SQLGenerator.class);
+            config.put("jndiDataSourceName", "jdbc/FileSyncJobTest");
+            config.put("database", TestUtil.DATABASE);
+            config.put("schema", TestUtil.SCHEMA);
+            dao.setConfig(config);
+
+            BucketMode bucketMode = FileSystemStorageAdapter.BucketMode.URI;
```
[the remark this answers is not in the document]
hjeeves: done

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: WriteException thrown, caught in outer try/catch for error log before end of run().

comment added for the method.

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
```diff
@@ -0,0 +1,229 @@
...
+        log.debug("cleaning stored artifacts...");
+        Iterator<Artifact> storedArtifacts = dao.storedIterator(null);
+        wipe_clean(storedArtifacts);
+
+        log.debug("cleaning unstored artifacts...");
+        Iterator<Artifact> unstoredArtifacts = dao.unstoredIterator(null);
+        wipe_clean(unstoredArtifacts);
+
+        // Clean up critwall tests in fsroot
+        log.debug("deleting test dir: " + TEST_ROOT);
```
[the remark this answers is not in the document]
hjeeves: Changed the comment, although I'm not convinced this shouldn't just be a Java File recursive function that deletes the entire directory. A 'gradle clean build' will blow the test directory away as well, but if that's not run between every intTest run, the main directory is likely to still be there. Tests run happily time after time, so it doesn't appear to be a problem under normal running state.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: Write Exception is now thrown as a fatal error.

PR #115, opened 2020-04-21
Fenwick main class

at88mph:


fenwick/src/main/java/org/opencadc/fenwick/Main.java
```diff
@@ -1,114 +1,247 @@
...
+                System.exit(2);
+            }
+
             InventoryHarvester doit = new InventoryHarvester(daoConfig, resourceID, selector);
             doit.run();
             System.exit(0);
         } catch (Throwable unexpected) {
-            log.error("unexpected failure", unexpected);
+            log.error("Unexpected failure", unexpected);
+            System.err.println("Unexpected failure.  Check log output.");
```
pdowler, line 179: Is System.err really going to be more visible than the logs, given that we run in a container and log to stdout?

fenwick/src/main/java/org/opencadc/fenwick/Main.java
```diff
@@ -1,114 +1,247 @@
...
-            
-            // parse config file and populate/assign these
-            Map<String,Object> daoConfig = new TreeMap<>();
-            URI resourceID = null;
-            ArtifactSelector selector = null;
-            
+            final MultiValuedProperties props = new MultiValuedProperties();
+
+            try {
+                Main.configure(props);
```
pdowler, line 115: this is done wrong. Reading config should be done with exactly two lines:
```
PropertiesReader pr = new PropertiesReader("fenwick.properties");
MultiValuedProperties props - pr.read();
```
Yes, this is wrong in tantar as well, but cadc-util has since been fixed to support the above and MVP now has useful single-valeud prop methods.

fenwick/src/main/java/org/opencadc/fenwick/Main.java
```diff
@@ -1,114 +1,247 @@
...
+     * Set any package's logging level here.  Allows for finer grained messages.
+     *
+     * @param applicationProperties The properties read from a file.
+     */
+    private static void setLogging(final MultiValuedProperties applicationProperties) {
+        Log4jInit.setLevel("ca.nrc.cadc", Level.WARN);
+        Log4jInit.setLevel("org.opencadc", Level.WARN);
+
+        final String loggingKey = ".logging";
+
```
pdowler, line 237: Why is there a loop here? The README says the value of the logging prop is a single "log level" value. We decided to hard code packages that get that level.

fenwick/src/main/java/org/opencadc/fenwick/Main.java
```diff
@@ -1,114 +1,247 @@
...
+            System.err.println("Unexpected failure.  Check log output.");
             System.exit(-1);
         }
     }
-    
-    private Main() { 
+
+    /**
+     * Read in the configuration file and load it into the given Properties.
+     */
```
pdowler, line 186: this method not needed -- applications don't encode the "find config file" logic

fenwick/src/main/java/org/opencadc/fenwick/Main.java
```diff
@@ -1,114 +1,247 @@
...
- * 
+ *
  * @author pdowler
  */
 public class Main {
+
     private static final Logger log = Logger.getLogger(Main.class);
+    private static final String CONFIG_FILE_NAME = "fenwick.properties";
+    private static final String CONFIG_PREFIX = Main.class.getPackage().getName();
+    private static final String SQLGENERATOR_CONFIG_KEY = SQLGenerator.class.getName();
```
pdowler, line 102: I find String.format to be quite hard to read compared to a single string + operator:
```
DB_SCHEMA_CONFIG_KEY = CONFIG_PREFIX + ".db.schema";
```
I think it has it's place (usually in building error messages) but seems to be more obscure for basic things.

fenwick/src/main/java/org/opencadc/fenwick/Main.java
at88mph: I assume you mean `getAllProperties()`.  I don't see a `read()` method.

at88mph:
Requested changes made.

fenwick/src/main/java/org/opencadc/fenwick/Main.java
pdowler: uh, yeah... that one :-)

PR #116, opened 2020-04-22
New IncludeArtifacts to include SQL files to specify Artifacts.

at88mph:


fenwick/src/main/java/org/opencadc/fenwick/ArtifactSelector.java
```diff
@@ -67,15 +67,35 @@
...
+ * Artifacts to be included in the metadata sync merge.
+ *
  * @author pdowler
  */
 public abstract class ArtifactSelector {
-    private static final Logger log = Logger.getLogger(ArtifactSelector.class);
+    final Path selectorConfigDir;
+
+    protected ArtifactSelector() {
+        // Read in the configuration file, and create a usable Map from it.
```
pdowler, line 85: Shouldn't all this be in the IncludeArtifacts impl?

If there is nothing  usefully common between ArtifactSelector impls, then this abstract class could be changed to an interface. When I stubbed this out we didn't know if that was the case or not..

fenwick/src/main/java/org/opencadc/fenwick/AllArtifacts.java
```diff
@@ -80,4 +84,15 @@
...
+
+    /**
+     * Obtain the iterator of clauses used to build a query to include the Artifacts being merged.
+     *
+     * @return An empty iterator.
+     */
+    @Override
+    public Iterator<String> iterateIncludeClauses() {
+        log.debug("No filters for AllArtifacts.");
+        return Collections.emptyIterator();
```
pdowler, line 96: I think this needs to return an iterator with a single no-op condition... 0-length string could be made to work but could easily be interpretted as a mistake elsewhere. Otherwise, artifact sync loop is going to have to special-case the empty iterator to do a single query, rather than just do a query for each clause returned by the iterator.

fenwick/src/main/java/org/opencadc/fenwick/ArtifactSelector.java
```diff
@@ -67,15 +67,35 @@
...
+        final File homeConfigDirectory = new File(String.format("%s/config/include", System.getProperty("user.home")));
 
-    protected ArtifactSelector() { 
+        // Allow some flexibility to override.  This is useful for local testing as one would normally require root
```
pdowler, line 88: Applications only have to support $HOME/config/include - the container setup is responsible for making sure /config is found in $HOME.

fenwick/src/main/java/org/opencadc/fenwick/AllArtifacts.java
```diff
@@ -80,4 +84,15 @@
     public AllArtifacts() {
         super();
     }
+
+    /**
+     * Obtain the iterator of clauses used to build a query to include the Artifacts being merged.
+     *
+     * @return An empty iterator.
+     */
+    @Override
```
pdowler, line 93: Unless we need multiple iterator-returning methods, just name this one `iterator()`

fenwick/src/main/java/org/opencadc/fenwick/IncludeArtifacts.java
```diff
@@ -0,0 +1,167 @@
...
+        if (configurationDirFile.isDirectory() && configurationDirFile.canRead()) {
+            final File[] fileListing = Objects.requireNonNull(
+                    configurationDirFile.listFiles(pathname -> pathname.getName().toLowerCase().endsWith(".sql")));
+            for (final File f : fileListing) {
+                boolean isFirstLine = true;
+                final StringBuilder clauseBuilder = new StringBuilder();
+                // There may be multiple lines per file.  Trim each one and treat it like a single clause.
+                for (final String clause : Files.readAllLines(f.toPath())) {
+                    if (StringUtil.hasText(clause)) {
+                        // The first line MUST begin with the WHERE keyword.
```
pdowler, line 137: I would allow for comment lines in these files (lines starting with --) so I guess you have to read them with a LineNumberReader and discard comments and I'd probably sanitise white space (at least blank lines and maybe newlines as well). At a minimum, we could then have an example that worked with an explanation.

All that stuff currently works fine in our TAP service but I don't think it is strictly required by TAP and it makes it hard to find the WHERE.

fenwick/src/main/java/org/opencadc/fenwick/ArtifactSelector.java
at88mph: OK.  Can I make it an interface then?  I thought there may be more implementations that would use this, but I suppose not.

fenwick/src/main/java/org/opencadc/fenwick/IncludeArtifacts.java
at88mph: Done.

at88mph:
Rework committed.

fenwick/src/main/java/org/opencadc/fenwick/AllArtifacts.java
pdowler: There is a change here, but returning the string "EOF" won't allow the sync loop to blindly pass that to the query generator. That should end up doing a query like this:
```
select * from inventory.Artifact where (EOF) and ...
```  
The goal is for a configuration with the AllArtifacts selector to execute a single query with no filtering on content. To do that without treating this selector as special, the iterator needs to return one item that can be included in the WHERE clause of the query without filtering out any artifacts. All these would work: `id is not null`, `uri is not null`, and 0-length (or whitespace) string, but I'm not sure which of those would be least/most confusing to a developer in the future. Ops people would never see it except if they look at the queries executed in luskan... then they might also get confused by some of them. 

If it was me and only I was ever going to look at the code, I'd just put a 0-length string right in the
```
return Collections.singletonList("").iterator();
```
and make sure that worked correctly... yeah, this is looking forward to implementing the sync loop and query generator, which will look something like:
```
Iterator<String> filters = selector.iterator();
while (filters.hasNext()) {
   String adql = generateADQL(filters.next());
   ResourceIterator<Artifact> ri = tapClient.execute(adql, artifactMapper);
   ...
}
```
To do that, AllArtifacts has to return one string that can be blindly put in the where clause of the ADQL query. 

**Other clarifications that would be good to add now:**

The ArtifactSelector javadoc for the iterator() method should specify exactly what is going to be done with each string. 

I would restrict the exceptions that are throwable from ArtifactSelector.iterator()... Maybe specify these standard exceptions for a few predictable cases - this is probably enough:
* ResourceNotFoundException when required configuration is missing
* IOException for unreadable configuration (eg impls can just let those fly)
* IllegalStateException for invalid configuration content
It's possible that if we were to allow for retrieving config from a URL that many/all the exceptions thrown by HttpTransfer.prepare() would be possible

fenwick/src/main/java/org/opencadc/fenwick/ArtifactSelector.java
pdowler: This was moved to IncludeArtifacts but it looks like this still defaults to /config/include -- we don't want that in code because the existence of /config is strictly a container operator thing. The current standard for our s/w is to look in $HOME/config only.

at88mph: Understood.  If it doesn't exist would we throw an Exception?

fenwick/src/main/java/org/opencadc/fenwick/AllArtifacts.java
at88mph: Thank you.  I thought there would be a special case for ignoring the SQL, such as EOF.

at88mph:
Thanks.  Rework submitted.

PR #121, opened 2020-04-28
bug fix for iterator auto commit state and change to use ResourceIterator

pdowler:


cadc-inventory-db/src/main/java/org/opencadc/inventory/db/SQLGenerator.java
```diff
@@ -1072,6 +1073,24 @@ public Object extractData(ResultSet rs) throws SQLException {
             this.con = con;
             this.rs = rs;
             hasRow = rs.next();
+            log.debug("ArtifactResultSetIterator: " + super.toString() + " ctor " + hasRow);
+            if (!hasRow) {
+                log.warn("ArtifactResultSetIterator:  " + super.toString() + " ctor - setAutoCommit(true)");
+                con.setAutoCommit(true);
```
at88mph, line 1079: This is assuming the connection's `autocommit` was `true` when the Iterator is entered.  The current state could be put aside and reset again.  Is it set to `false` somewhere?  I don't see why setting the `autocommit` flag when exiting is needed.  Insurance?

pdowler: It is set to false up on line 572 right before the ArtifactResultSetIterator is constructed; this line is needed so that an empty result set (next() returns false) does not leave the connection with an open txn. That was the subtle bug that caused hangs depending on sequence of events.

It is admittedly not well organised with the auto commit false and true in different parts of the code. I would have to combine the two nested classes to fix it.

PR #123, opened 2020-04-29
s2575, ta13293: FileSyncJob only works on URLs that match artifact data

hjeeves:


critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -241,41 +244,59 @@ public void run() {
...
+
+                try {
+                    ByteArrayOutputStream dest = new ByteArrayOutputStream();
+                    HttpGet get = new HttpGet(u, dest);
+                    get.prepare();
+
+                    // Check content checksum and length to verify this
+                    // artifact can be synced currently
+                    Artifact curArtifact = artDao.get(artifact);
+                    log.debug("content checksums: " + get.getContentMD5() + "\n" + curArtifact.getContentChecksum().getSchemeSpecificPart());
```
pdowler, line 265: These checks here will work out if the remote server provides the values in the header metadata, but they are optional so get.getContentMD5() and get.getContentLength() can return nulls. The same methods in curArtifact never return null (see data model).

So these checks need to be tweaked to avoid NullPointerException.

Whether or not the `get` provides values for contentMD5 or contentLength, the values from curArtifact ***must*** be included with `NewArtifact a` below so that the storage adapter can verify the downloaded data actually matches the expectation. That protects against corrupted downloads and as well as the case where one of the sites has a different copy of the file (either newer or older - no way to tell here).

Technically, you could not do any checks here and rely on the storage adapter and it would get the correct outcome. However, the adapter would have to download the whole file in order to fail so the checks here are a optimisation. Please add a comment to that effect.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -241,41 +244,59 @@ public void run() {
...
+                log.debug("trying " + u);
+
+                try {
+                    ByteArrayOutputStream dest = new ByteArrayOutputStream();
+                    HttpGet get = new HttpGet(u, dest);
+                    get.prepare();
+
+                    // Check content checksum and length to verify this
+                    // artifact can be synced currently
+                    Artifact curArtifact = artDao.get(artifact);
```
pdowler, line 264: It is a good thing to retrieve the Artifact from the database here because in principle it's state could have changed since the iterator returned the row and a FileSyncJob was created. It wasn't specified in the task, but at this point there should be a check that the artifact is still a valid thing to sync: if it now has a storageLocation, the fileSyncJob should stop/return with no further action.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -241,41 +244,59 @@ public void run() {
...
+                    log.info("write error for storage adapter put: " + wre.getMessage());
+                    throw wre;
+                } catch (TransientException | IOException te) {
+                    // ReadException will be caught under the IOException
+                    // - prepare or put throwing this error
+                    // - will move to next url
+                    log.info("transient exception." + te.getMessage());
+
+                } catch (ResourceNotFoundException | ResourceAlreadyExistsException | PreconditionFailedException badURL) {
+                    log.info("removing URL " + u + " from list: " + badURL.getMessage());
```
pdowler, line 291: PreconditionFailedException probably means the URL should be discarded. With the code above it should be, but if the PreconditionFailedException was thrown by StorageAdapter.put it could have been a corrupt download that didn't trigger a ReadException (premature EOF maybe)... still, let's stick with discarding the URL for now.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -241,41 +244,59 @@ public void run() {
      * @throws InterruptedException
      * @throws WriteException
      */
-    private StorageMetadata syncArtifact(Iterator<URL> urlIterator, URI artifact, StorageAdapter sa)
-        throws StorageEngageException, InterruptedException, WriteException {
+    private StorageMetadata syncArtifact(Iterator<URL> urlIterator, URI artifact, StorageAdapter sa, ArtifactDAO artDao)
```
pdowler, line 247: I didn't notice this before, but it's confusing to pass member variables as args to methods: the method can use the member var directly. Further, these 3 are all final specifically so methods cannot accidentally set the values to something else (like null). Only the iterator is needed).

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
```diff
@@ -195,31 +212,111 @@ public void testValidJob() {
...
+            createTestDirectory(testDir);
+
+            OpaqueFileSystemStorageAdapter sa = new OpaqueFileSystemStorageAdapter(new File(TEST_ROOT), 1);
+
+            URI artifactID = new URI(TEST_ARTIFACT_URI);
+            URI resourceID = new URI(TEST_RESOURCE_ID);
+
+            // Set up an Artifact in the database to start.
+            // Set checksum to the wrong value.
+            Artifact artifactToUpdate = new Artifact(artifactID,
```
pdowler, line 255: Can you make the bad checksum more obviously incorrect here? When I looked, it started the same as the correct one and it took some time to see the discrepancy. Instead of:
```
md5:646d3c548ffb98244a0fc52b6055555
```
maybe this looks more made-up:
```
md5:0123456789abcdef0123456789abcdef
```
note:  used sequence so it's easier to get the right length

hjeeves: ok

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
hjeeves: ok

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -233,49 +236,83 @@ public void run() {
...
+                    HttpGet get = new HttpGet(u, dest);
+                    get.prepare();
 
+                    // Check content checksum and length to verify this
+                    // artifact can be synced currently
+                    Artifact curArtifact = this.artifactDAO.get(this.artifactID);
+
+                    if (curArtifact.storageLocation != null) {
+                        // artifact has been acted on since this FileSyncJob
+                        // instance started and it was null
```
pdowler, line 267: I would have agonised about what exception to throw here. I'm generally reluctant to use the exception mechanism for flow control but we already used null to mean "nothing accomplished but not fatal either" and this scenario does need to terminate the job.

`IllegalArgumentException` typically means the input (iterator) is invalid input for this method and could be thrown by other code in here (because it gets thrown a lot by constructors). 

Since the input was fine and the scenario is that the state of the system as a whole is not as expected, I would have ended up chosing `IllegalStateException` and had a separate catch for it in run() -- because when we go back through here and make it log in json format for ingestion  into ELK, we'll have to distinguish whether this "job" was success = true|false and I would want that logged as success=true when this happens. 

It's admittedly seems pretty pedantic :-)

hjeeves: I did agonise to a point - but also knew you'd tell me if I got it wrong. :)

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
```diff
@@ -195,31 +213,157 @@ public void testValidJob() {
...
+                new URI("md5:646d3c548ffb98244a0fc52b60556082"), new Date(),
+                1008000L);
+            StorageLocation testLocation = new StorageLocation(artifactID);
+            artifactToUpdate.storageLocation = testLocation;
```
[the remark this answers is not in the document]
hjeeves: hopefully that's got it now.

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
```diff
@@ -195,31 +213,157 @@ public void testValidJob() {
...
+        try {
+            createTestDirectory(testDir);
+
+            OpaqueFileSystemStorageAdapter sa = new OpaqueFileSystemStorageAdapter(new File(TEST_ROOT), 1);
+
+            URI artifactID = new URI(TEST_ARTIFACT_URI);
+            URI resourceID = new URI(TEST_RESOURCE_ID);
+
+            // Set up an Artifact in the database to start.
+            // Set checksum to the wrong value.
```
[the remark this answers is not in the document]
hjeeves: ok

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
```diff
@@ -195,31 +213,157 @@ public void testValidJob() {
...
+        } catch (Exception unexpected) {
+            log.debug("unexpected exception: " + unexpected);
+            Assert.fail("unexpected exception");
+        }
+
+        log.info("testInvalidJobBadContentLen - DONE");
+    }
+
+    @Test
+    public void testInvalidJobBadStorageLocation() {
```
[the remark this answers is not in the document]
hjeeves: test name changed

PR #127, opened 2020-04-30
Fenwick site sync and add trackSiteLocations property to enable it

at88mph:


fenwick/src/main/java/org/opencadc/fenwick/Main.java
```diff
@@ -170,7 +166,12 @@ public static void main(final String[] args) {
                 System.exit(2);
             }
 
-            InventoryHarvester doit = new InventoryHarvester(daoConfig, resourceID, selector);
+            final String configuredTrackSiteLocations = Main.readProperty(props, TRACK_SITE_LOCATIONS_CONFIG_KEY,
+                                                                          null);
+            final boolean trackSiteLocations = StringUtil.hasText(configuredTrackSiteLocations)
```
pdowler, line 171: The way this is handled, trackSiteLocations is optional and defaults to false? I had been trying to maintain a policy of either "configured" or "hard-coded" but no "optional-with-default"... probably I didn't make that clear (it isn't written down). I find "optional-with-default" generally leads to unexpected and therefore wrong behaviour. In this case, one could waste a lot of time and resources building global inventory that didn't know where any files where :-)

Please review the config parsing and make every item required.

at88mph:
I did not know that the sites would be stored in the local inventory database.  This is added.  Does the `StorageSiteDAO` PUT duplicates?

at88mph:
Committed now.

fenwick/src/main/java/org/opencadc/fenwick/StorageSiteSync.java
```diff
@@ -0,0 +1,204 @@
...
+                throw new IllegalStateException("No storage sites available to sync.");
+            }
+        }
+    }
+
+    /**
+     * Store the StorageSite in the local inventory database.
+     * @param storageSite   The StorageSite to store.
+     */
+    void putStorageSite(final StorageSite storageSite) {
```
pdowler, line 157: This is a lot of code for a 1-line autocommit transaction. You only need explicit transaction handling if you are doing more than one change via DAO classes and need it to be atomic.

Please just inline the single line up at line~144.

fenwick/src/main/java/org/opencadc/fenwick/Main.java
```diff
@@ -162,6 +159,18 @@ public static void main(final String[] args) {
                                               ? InventoryUtil.loadPlugin(configuredArtifactSelector)
                                               : null;
 
+            final String configuredTrackSiteLocations = Main.readProperty(props, TRACK_SITE_LOCATIONS_CONFIG_KEY,
+                                                                          errorMessage);
+            final boolean trackSiteLocations;
+            if (StringUtil.hasText(configuredTrackSiteLocations)) {
+                trackSiteLocations = StringUtil.hasText(configuredTrackSiteLocations)
+                                     && Boolean.parseBoolean(configuredTrackSiteLocations);
+            } else {
```
pdowler, line 168: The comment and assignment here just make defaulting to false more explicit. I asked for "review the config parsing and make every item required". If trackSiteLocations (or anything else!) is not configured, fenwick should fail.

fenwick/src/main/java/org/opencadc/fenwick/StorageSiteSync.java
```diff
@@ -0,0 +1,204 @@
...
+    public StorageSite doit() throws AccessControlException, ResourceNotFoundException,
+                                     ByteLimitExceededException, NotAuthenticatedException,
+                                     IllegalArgumentException, TransientException, IOException,
+                                     InterruptedException {
+        try (final ResourceIterator<StorageSite> storageSiteIterator = queryStorageSites()) {
+            if (storageSiteIterator.hasNext()) {
+                final StorageSite storageSite = storageSiteIterator.next();
+                if (storageSiteIterator.hasNext()) {
+                    throw new IllegalStateException("More than one Storage Site found.");
+                } else {
```
pdowler, line 143: Before putting the StorageSite locally, you need to verify that the metaChecksum in the returned StorageSite's metaChecksum is correct by computing a new value and comparing to the assigned value. This method must fail on a mismatch.

the `storageSiteDAO.put()` and return do not need to be inside an else here

fenwick/src/main/java/org/opencadc/fenwick/StorageSiteSync.java
```diff
@@ -0,0 +1,204 @@
...
+
+
+/**
+ * Class to pull a single StorageSite from a remote TAP (Luskan) service.  This class is enabled by the
+ * org.opencadc.fenwick.trackSiteLocations property in the fenwick.properties file.
+ */
+public class StorageSiteSync {
+    private static final Logger log = Logger.getLogger(StorageSiteSync.class);
+
+    private static final String STORAGE_SITE_QUERY = "SELECT name, id, resourceid FROM inventory.storagesite";
```
pdowler, line 104: You also need to get the lastModified and metaChecksum columns in the query...

fenwick/src/main/java/org/opencadc/fenwick/StorageSiteSync.java
```diff
@@ -0,0 +1,204 @@
...
+     * @throws ResourceNotFoundException remote resource not found
+     * @throws TransientException temporary failure of TAP service: same call could work in future
+     * @throws IOException failure to send or read data stream
+     * @throws InterruptedException thread interrupted
+     */
+    ResourceIterator<StorageSite> queryStorageSites() throws AccessControlException, ResourceNotFoundException,
+                                                             ByteLimitExceededException, NotAuthenticatedException,
+                                                             IllegalArgumentException, TransientException, IOException,
+                                                             InterruptedException {
+        return tapClient.execute(STORAGE_SITE_QUERY, row -> {
```
pdowler, line 197: This also needs to assign the entity fields (lastModified and metaChecksum) using InventoryUtil.asssignLastModified and InventoryUtil.assignMetaChecksum respectively.

fenwick/src/test/java/org/opencadc/fenwick/StorageSiteSyncTest.java
```diff
@@ -0,0 +1,248 @@
...
+import org.junit.Test;
+import org.opencadc.inventory.StorageSite;
+import org.opencadc.tap.TapClient;
+
+
+public class StorageSiteSyncTest {
+    static {
+        Log4jInit.setLevel("org.opencadc.fenwick", Level.DEBUG);
+    }
+
```
pdowler, line 92: Admittedly, it isn't easy to write tests for this kind of code, but the following overrides so much that there is more test code than the real code in the doit() method. 

Effort might be better spent on intTest: 
* the TAP query and TapRowMapper test would need a locally deployed minoc and luskan (and their common db), plus a reg for lookups
* the actual sync in doit() would require an additonal destination db
These will be needed to test various parts of fenwick

fenwick/src/main/java/org/opencadc/fenwick/StorageSiteSync.java
at88mph: This makes sense, thanks.

fenwick/src/main/java/org/opencadc/fenwick/Main.java
at88mph: Alright.  The logic here was borrowed from Critwall, where each missing configuration is gathered and reported at the end.  Is it preferred to have one pass through and check all configuration, and then pull the values out?  That may be more explicit.

fenwick/src/main/java/org/opencadc/fenwick/StorageSiteSync.java
at88mph: OK.  Does the `Entity` class need constructor args for those values?  There is no way to set them currently.

fenwick/src/test/java/org/opencadc/fenwick/StorageSiteSyncTest.java
at88mph: I'll make integration tests a task then and get those written.

fenwick/src/main/java/org/opencadc/fenwick/StorageSiteSync.java
at88mph: Nevermind, I missed the next comment.

fenwick/src/main/java/org/opencadc/fenwick/Main.java
at88mph: I've reworked this.  Please look.

at88mph:
The StorageSiteSync class has been modified to add the `lastmodified` and `metachecksum` column values.  It will also verify the `metachecksum` value of the site before storing it locally.  The `put` does NOT issue a `forceUpdate` flag.

Working on integration testing now.

at88mph:
I've added an integration test now.  For error conditions, the query iterator is still being overridden unless there's a better way, which I can only think of multiple sites to use, or modifying the site the test is using to change the checksum or something similar.

fenwick/src/intTest/java/org/opencadc/fenwick/StorageSiteSyncTest.java
```diff
@@ -0,0 +1,373 @@
...
+     * Test StorageSite synchronization.
+     * Test one: Populate the local inventory database with a StorageSite from the remote Luskan.
+     * Test two: Ensure the same site is put again without error.
+     * Test three: Ensure a failed execution if that same site has a different checksum than its computed one.
+     * Test four: Ensure a failed execution if multiple sites are found.
+     * Test five: Ensure a failed execution if no sites are found.
+     *
+     * @throws Exception
+     */
+    @Test
```
pdowler, line 181: This needs to be broken up into separate test methods so it is easier to run some of the tests (eg when diagnosing an issue) and so the cursory summary printed by "gradle intTest" tells you what worked and what didn't -- if everything fails it is usually config.. A single method can do multiple things (like one and two since they build on each other) ... but each kind of fail should be it's own test.

fenwick/src/test/java/org/opencadc/fenwick/StorageSiteSyncTest.java
```diff
@@ -0,0 +1,274 @@
...
+                };
+            }
+        };
+
+        final StorageSite storageSiteResult = testSubject.doit();
+
+        Assert.assertEquals("Wrong site list.", storageSiteList.get(0), storageSiteResult);
+    }
+
+    @Test
```
pdowler, line 143: This test looks useful to keep - maybe rename to testDetectMultiple

fenwick/src/test/java/org/opencadc/fenwick/StorageSiteSyncTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+        try {
+            testSubject.doit();
+            Assert.fail("Should throw IllegalStateException");
+        } catch (IllegalStateException e) {
+            // Good.
+            Assert.assertEquals("Wrong message.", "More than one Storage Site found.", e.getMessage());
+        }
+
+        // Second kind of exception
```
pdowler, line 185: split this test out into a separate method - testStorageSiteNotFound

fenwick/src/test/java/org/opencadc/fenwick/StorageSiteSyncTest.java
pdowler: I would remove the currently skipped (@Ignore) methods - too much mocking to really be useful.

at88mph:
Thanks.  Rework is complete but compilation will fail until the URI casting from the TapRowMapper is fixed.

PR #129, opened 2020-05-08
s2571 - initial commit of DeletedStorageLocationEventSync

jburke-cadc:


fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
```diff
@@ -0,0 +1,132 @@
...
+import org.opencadc.inventory.DeletedStorageLocationEvent;
+import org.opencadc.inventory.InventoryUtil;
+import org.opencadc.tap.TapClient;
+import org.opencadc.tap.TapRowMapper;
+
+public class DeletedStorageLocationEventSync {
+
+    private static final Logger log = Logger.getLogger(DeletedStorageLocationEventSync.class);
+
+    // TODO sufficient to use 3 digit precision ISO_DATE_FORMAT in DateUtil?
```
pdowler, line 92: yes, use the DateUtil constant

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
```diff
@@ -0,0 +1,132 @@
...
+    public static final String ISO_DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSSSSS";
+    private static final DateFormat dateFormat = DateUtil.getDateFormat(ISO_DATE_FORMAT, DateUtil.UTC);
+
+    private static final String DELETED_STORAGE_LOCATION_QUERY =
+            "SELECT id, lastModified, metaChecksum FROM inventory.DeletedStorageLocationEvent WHERE lastModified >= %s";
```
pdowler, line 97: The query has to` order by lastModified` in order to incrementally move forward and for the returned iterator to havethe correct ordering

fenwick/src/main/java/org/opencadc/fenwick/DeletedArtifactEventSync.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class DeletedArtifactEventSync {
+
+    private static final Logger log = Logger.getLogger(DeletedArtifactEventSync.class);
+
+    private static final DateFormat dateFormat = DateUtil.getDateFormat(DateUtil.ISO_DATE_FORMAT, DateUtil.UTC);
+
+    private static final String DELETED_ARTIFACT_QUERY =
+        "SELECT id, lastModified, metaChecksum FROM inventory.DeletedArtifactEvent "
+            + "WHERE lastModified >= %s order by lastModified";
```
pdowler, line 96: The query has to take into account that startTime might be null (on first run)

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
```diff
@@ -0,0 +1,131 @@
...
+
+public class DeletedStorageLocationEventSync {
+
+    private static final Logger log = Logger.getLogger(DeletedStorageLocationEventSync.class);
+
+    private static final DateFormat dateFormat = DateUtil.getDateFormat(DateUtil.ISO_DATE_FORMAT, DateUtil.UTC);
+
+    private static final String DELETED_STORAGE_LOCATION_QUERY =
+            "SELECT id, lastModified, metaChecksum FROM inventory.DeletedStorageLocationEvent "
+                + "WHERE lastModified >= %s order by lastModified";
```
pdowler, line 96: The query has to take into account that startTime might be null (on first run)

at88mph: The `%s` date comparison needs to be wrapped in single quotes, too.

jburke-cadc: Can you try pulling this branch again. The last pull request should fix this.

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.log4j.Logger;
+import org.opencadc.inventory.DeletedStorageLocationEvent;
+import org.opencadc.inventory.InventoryUtil;
+import org.opencadc.tap.TapClient;
+import org.opencadc.tap.TapRowMapper;
+
+public class DeletedStorageLocationEventSync {
+
+    private static final Logger log = Logger.getLogger(DeletedStorageLocationEventSync.class);
+
```
pdowler, line 91: DateFormat (and Calendar) classes are not thread safe so static instances should never be used (on principle). Just having a member var makes DeletedStorageLocationEventSync non-thread safe, which is fine as long as you know that is it called by one thread only... but I usually create them as locally as possible (here, that would be as a member of the TapRowMapper instance)

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
```diff
@@ -0,0 +1,131 @@
...
+    private static final DateFormat dateFormat = DateUtil.getDateFormat(DateUtil.ISO_DATE_FORMAT, DateUtil.UTC);
+
+    private static final String DELETED_STORAGE_LOCATION_QUERY =
+            "SELECT id, lastModified, metaChecksum FROM inventory.DeletedStorageLocationEvent "
+                + "WHERE lastModified is not null AND lastModified >= '%s' order by lastModified";
```
pdowler, line 96: The lastModified value in the database is never null. The startTime value you are going to use when executing the query is going to be null the first time events are synced from a specific source.

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private final TapClient<DeletedStorageLocationEvent> tapClient;
+    private final Date startTime;
```
pdowler, line 99: startTime can be null, so it should be public and not final

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public DeletedStorageLocationEventSync(TapClient<DeletedStorageLocationEvent> tapClient, Date startTime) {
+        this.tapClient = tapClient;
+        this.startTime = startTime;
```
pdowler, line 103: startTime is optional: not set in constructor

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
jburke-cadc: Moved creating the DateFormat into an instance method. If there's no likelihood of the DateFormat method being overridden, could move it directly into the query building.

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
jburke-cadc: Made startTime public

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
jburke-cadc: Removed from the constructor.

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
jburke-cadc: Got that backwards. Fixed the query.

fenwick/src/main/java/org/opencadc/fenwick/DeletedArtifactEventSync.java
```diff
@@ -0,0 +1,165 @@
...
+     *
+     * @return Iterator over the query results.
+     * @throws InterruptedException thread interrupted
+     * @throws IOException failure to write or read the data stream
+     * @throws ResourceNotFoundException remote resource not found
+     * @throws TransientException temporary failure of TAP service: same call could work in future
+     */
+    public ResourceIterator<DeletedArtifactEvent> getEvents()
+        throws InterruptedException, IOException, ResourceNotFoundException, TransientException {
+
```
pdowler, line 124: This use of String.format for building a query is more complicated than the problem. Just use a StringBuilder and `append` the parts so it is readable in one place.

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
```diff
@@ -0,0 +1,167 @@
...
+     *
+     * @return Iterator over the query results.
+     * @throws InterruptedException thread interrupted
+     * @throws IOException failure to write or read the data stream
+     * @throws ResourceNotFoundException remote resource not found
+     * @throws TransientException temporary failure of TAP service: same call could work in future
+     */
+    public ResourceIterator<DeletedStorageLocationEvent> getEvents()
+            throws InterruptedException, IOException, ResourceNotFoundException, TransientException {
+
```
pdowler, line 124: same comment as for DeletedArtifactSync

fenwick/src/main/java/org/opencadc/fenwick/DeletedArtifactEventSync.java
jburke-cadc: Updated to use a StringBuilder to create the query.

fenwick/src/main/java/org/opencadc/fenwick/DeletedStorageLocationEventSync.java
jburke-cadc: As above.

PR #130, opened 2020-05-12
Fenwick artifact sync

at88mph:


fenwick/src/main/java/org/opencadc/fenwick/ArtifactSync.java
```diff
@@ -0,0 +1,184 @@
...
+     * @return String ADQL query with expected WHERE clauses.
+     *
+     * @throws ResourceNotFoundException For any missing required configuration that is missing.
+     * @throws IOException               For unreadable configuration files.
+     * @throws IllegalStateException     For any invalid configuration.
+     */
+    private String assembleQuery() throws ResourceNotFoundException, IOException, IllegalStateException {
+        final StringBuilder whereClauseBuilder = new StringBuilder();
+        final Iterator<String> whereClauseIterator = selector.iterator();
+
```
pdowler, line 178: Each string in the iterator is from a different include file so defined a different filter... they cannot be added together with AND becaue they most likely select different sets of artifacts. In the query here, you need to have only one of those.

and that whole clause has to be  in `( )` in case it has `OR` in it

at88mph:
The `ArtifactSelector` may not suit the role that it was intended to other than to read in the constraint SQL files.  The Luskan query already contains a default `WHERE` clause for the `lastModified` column, so the `AllArtifacts` could simply become a query that returns everything from the Epoch (or something similar).

The `ArtifactSync` method could have two `iterator()` methods:

```java
public ResourceIterator<Artifact> iterator(final String constraint) {}
```

and 

```java
public ResourceIterator<Artifact> iterator() {}
```

With the latter being a "query all artifacts" method.  The API would allow the constraint to be optional and obvious to anyone using it.  That way the `constraint` in the former method would be a `@NotNull` parameter to clearly indicate intent.

at88mph:
The requested changes are done except for the include file changes.  Can I proceed with the suggested ones?  Shall I just do the quick version of what you mentioned?

pdowler:
I was thinking just have ArtifactSelector return a List<String> so that no constraints (empty list) is easy to detect. It still requires special handling of the second and subsequent constraints but at least the main cases of "no constraints" and "one constraint" are straightforward and the intent to filter or not is still specified explicitly in the config.

I have changed the ArtifcatSelector API and effected code... creating PR now.

at88mph:
Integrating now.

at88mph:
Can the single constraint be passed as an argument to the `iterator()` method?  Is there a reason to add it as a public field instead?

at88mph:
I've committed a pass at it to get the ball rolling with the requested changes and a modified `iterator()` method.  Let me know, please.

at88mph:
Sorry if you're in the middle of looking at this.  I've changed ArtifactSync to have a mutable field as it was cleaner to keep consistent with the other objects in the storage-inventory model.

PR #136, opened 2020-05-25
S2655 - make the grant expiry time configurable

jburke-cadc:


baldur/src/intTest/java/org/opencadc/baldur/GetPermissionsTest.java
```diff
@@ -229,17 +230,10 @@ public Object run() throws Exception {
...
-                    
-                    GrantReader grantReader = new GrantReader();
-                    Grant grant = grantReader.read(out.toString());
-                    Assert.assertTrue(grant instanceof ReadGrant);
-                    ReadGrant readGrant = (ReadGrant) grant;
-                    Assert.assertEquals(URI.create(notFoundArtifactURL), readGrant.getArtifactURI());
-                    Assert.assertFalse(readGrant.isAnonymousAccess());
-                    Assert.assertTrue(readGrant.getGroups().isEmpty());
-                    
+                    Assert.assertEquals(404, get.getResponseCode());
```
pdowler, line 233: Please change to use HttpGet anywhere that is not downloading a file to local disk.

Checking the type of exception is only verifying the client behaviour. The response code 404 is sufficient for correct service behaviour.

baldur/README.md
```diff
@@ -10,6 +10,9 @@ The following configuration files must be available in the /config directory.
 
 The configuration in baldur.properties serves two purposes:  to allow certain users access to the service and to describe the permission rules for URIs that baldur will produce through its REST API.
 ```
+# time (in seconds) the grant is considered valid
```
pdowler, line 13: For consistency with other service config, the property keys should be 
```
org.opencadc.baldur.grantExpiry = {time in seconds}
org.opencadc.baldur.allowedUser = {user identity}
```
and specify that multiple occurances of allowedUser is supported. For consistency with other places (eg. logControl) where we configure usage permissions like this, I would also add:
```
org.opencadc.baldur.allowedGroup = {groupURI}
```
also multiple occurances. Then add "TODO: add support for allowedGroup" and create an issue so it's clear that it isn't supported but this is how it will be eventually. (see raven README and writeGrantProvider for how it is expressed there).

Instead of entry:
```
org.opencadc.baldur.entry = {entry name} {regular expression}
```

baldur/src/main/java/org/opencadc/baldur/PermissionsConfig.java
```diff
@@ -198,11 +201,34 @@ private void init() {
...
+            // get the Grant timeout
+            List<String> timeout = allProps.getProperty(KEY_EXPIRY_TIME);
+            if (timeout.size() == 0) {
+                throw new IllegalStateException("no values for key " + KEY_EXPIRY_TIME + " in " + PERMISSIONS_PROPERTIES);
+            }
+            if (timeout.size() > 1) {
+                throw new IllegalStateException("multiple values for key " + KEY_EXPIRY_TIME + " in "
+                    + PERMISSIONS_PROPERTIES + " where single value expected");
+            }
+            try {
```
pdowler, line 214: When used on thwe left-hand side of an assignment, I like to see
```
this.expiryDate =
```
so that when looking at code without an IDE I can tell that's a member variable assignment.

baldur/README.md
jburke-cadc: Instead of a single line for user or groups with space delimited entries, multiple lines for each. Okay that makes it easier to read.
Was there something more about ...baldur.entry ?

baldur/src/intTest/java/org/opencadc/baldur/GetPermissionsTest.java
jburke-cadc: Updated to use HttpGet.

baldur/README.md
pdowler: Comment updated to clarify `entry` -> `org.opencadc.baldur.entry`

baldur/src/intTest/java/org/opencadc/baldur/GetPermissionsTest.java
jburke-cadc: Updated to properly use HttpGet.

baldur/README.md
jburke-cadc: Renamed the property keys and created an issue to add support for allowdGroup.

baldur/src/main/java/org/opencadc/baldur/PermissionsConfig.java
jburke-cadc: The member variables in PermissionsConfig are static.

pdowler: ugh. I had not noticed that earlier. That severely limits how people can deploy instances of this and while we are currently targetting a container deployment so only one application per JVM the code could be deployed in other ways. Please change these to non-static.

jburke-cadc: Updated the member variables to be non-static. I did have to remove the caching that made use of the static variables. Do we need a new caching method?

PR #142, opened 2020-05-29
s2573 - update luskan availability error reporting to return a more detailed message...

jburke-cadc:
… of what went wrong.

### DIFF
--- a/luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
+++ b/luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
@@ -69,6 +69,11 @@ package org.opencadc.luskan.ws;
 
 import ca.nrc.cadc.auth.AuthMethod;
 import ca.nrc.cadc.db.DBUtil;
+import ca.nrc.cadc.net.ResourceNotFoundException;
+import ca.nrc.cadc.reg.AccessURL;
+import ca.nrc.cadc.reg.Capabilities;
+import ca.nrc.cadc.reg.Capability;
+import ca.nrc.cadc.reg.Interface;
 import ca.nrc.cadc.reg.Standards;
 import ca.nrc.cadc.reg.client.LocalAuthority;
 import ca.nrc.cadc.reg.client.RegistryClient;
@@ -84,7 +89,9 @@ import ca.nrc.cadc.vosi.avail.CheckResource;
 import ca.nrc.cadc.vosi.avail.CheckWebService;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.URI;
+import java.util.NoSuchElementException;
 import javax.sql.DataSource;
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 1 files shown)


luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
```diff
@@ -148,38 +155,85 @@ public AvailabilityStatus getStatus() {
...
+            }
+            log.error(sb.toString());
             isGood = false;
-            note = "test failed, reason: " + t;
+            note = sb.toString();
         }
         return new AvailabilityStatus(isGood, null, null, null, note);
     }
 
+    private String getServiceUrl(URI standardID, URI standard, AuthMethod authMethod)
```
pdowler, line 190: This is a vague method name with vague argument names for a very specialised bit of logic. How about `getAvailabilityForLocal(URI standardID)`. Note that LocalAuthority is a special way to lookup the local provider of some feature where it is inherent that there is one instance (so just A&A, not things like "the local TAP service").

luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
```diff
@@ -148,38 +155,85 @@ public AvailabilityStatus getStatus() {
...
+
+        Capabilities capabilities;
+        try {
+            capabilities = reg.getCapabilities(resourceID);
+        } catch (IOException e) {
+            String message = String.format("io error getting capabilities for resourceID %s, standardID %s, because %s",
+                                           resourceID, standardID, e.getMessage());
+            throw new ResourceNotFoundException(message);
+        }
+
```
pdowler, line 211: From here down, `ResourceNotFoundException` is not the correct error if a service does not implement the standard you are interested in nor if it does not support a specific AuthMethod (the latter especially is not a resource in any sense). 

Yes, this is captured above and not thrown as it would be in a library, but things like this tend to be copied/moved into libraries where choice of exception becomes entrenched.

Probably `UnsupportedOperationException("service %s does not provide %s", ...)` would work for findCapability. I don't know about findInterface(AuthMethod)... it's not an operation and is never going to happen in practice so I'd probably just go with `RuntimeException("unexpected: service %s capability %s does not support auth: %s", ...)`

luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
jburke-cadc: Renamed and updated the method

luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
jburke-cadc: Updated the error messages, thanks, much better.

luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
```diff
@@ -148,38 +155,86 @@ public AvailabilityStatus getStatus() {
...
+        Capability capability = capabilities.findCapability(Standards.VOSI_AVAILABILITY);
+        if (capability == null) {
+            String message =
+                String.format("service %s not does not provide %s", standardID, Standards.VOSI_AVAILABILITY);
```
pdowler, line 215: the first arg above should be the resourceID  found eariler in order to be helpful, not the standardID of the feature, same change below

luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
```diff
@@ -148,38 +155,86 @@ public AvailabilityStatus getStatus() {
...
+
+        Interface anInterface = capability.findInterface(AuthMethod.ANON);
+        if (anInterface == null) {
+            String message = String.format(
+                "unexpected: service %s capability %s does not support auth: %s",
+                standardID, capability, AuthMethod.ANON);
+            throw new RuntimeException(message);
+        }
+
+        AccessURL accessURL = anInterface.getAccessURL();
```
pdowler, line 227: The accessURL cannot be null so just return it here.

luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
jburke-cadc: Updated

luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
jburke-cadc: Updated.

PR #151, opened 2020-06-15
Test fixes and add isValid() check to protect from NPE.

at88mph:
- Added `isValid()` call to `ResolutionPolicy` comparison
 - Unit test fixes to adapt to URI requirements.


at88mph:
Here is the exception when putting `file.txt` manually into `1/2/3/4/5` and running Tantar.  It's not a `NullPointerException`, but it is an _accurate_ exception:

```
java.lang.RuntimeException: failed to recreate StorageMetadata: 12345,uuid:file.txt
	at org.opencadc.inventory.storage.fs.OpaqueIterator.createStorageMetadata(OpaqueIterator.java:224)
	at org.opencadc.inventory.storage.fs.OpaqueIterator.advance(OpaqueIterator.java:153)
	at org.opencadc.inventory.storage.fs.OpaqueIterator.<init>(OpaqueIterator.java:117)
	at org.opencadc.inventory.storage.fs.OpaqueFileSystemStorageAdapter.iterator(OpaqueFileSystemStorageAdapter.java:507)
	at org.opencadc.tantar.BucketValidator.lambda$iterateStorage$0(BucketValidator.java:604)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.opencadc.tantar.BucketValidator.iterateStorage(BucketValidator.java:602)
	at org.opencadc.tantar.BucketValidator.validate(BucketValidator.java:259)
	at org.opencadc.tantar.Main.main(Main.java:121)
Caused by: java.nio.file.FileSystemException: /var/local/data/content/1/2/3/4/5/file.txt: Unable to get size of extended attribute 'contentChecksum': No data available
	at sun.nio.fs.LinuxUserDefinedFileAttributeView.size(LinuxUserDefinedFileAttributeView.java:143)
	at org.opencadc.inventory.storage.fs.OpaqueFileSystemStorageAdapter.getFileAttribute(OpaqueFileSystemStorageAdapter.java:601)
	at org.opencadc.inventory.storage.fs.OpaqueIterator.createStorageMetadata(OpaqueIterator.java:209)
```

So in this case, the `isValid()` needs to be called by the Policy again to do something different, which I guess in this case would be to simply remove the file.

at88mph:
Rework added for handling case where `isValid()` returns `false`.

pdowler:
Oh, the FS adapter doesn't catch that correctly... fixing to correctly return an invalid instead.

at88mph:
I was just writing that...

at88mph:
I just added another check for the case where the Artifact doesn't already exist.  This should be ready for review now.

tantar/src/main/java/org/opencadc/tantar/policy/InventoryIsAlwaysRight.java
```diff
@@ -95,25 +95,31 @@ public void resolve(final Artifact artifact, final StorageMetadata storageMetada
...
+            reporter.report("Removing Unknown File " + storageMetadata.getStorageLocation() + " as per policy.");
 
             validateEventListener.delete(storageMetadata);
         } else {
             // Check metadata for discrepancies.
             if (haveDifferentStructure(artifact, storageMetadata)) {
-                // Then prefer the Artifact.
-                reporter.report(String.format("Replacing File %s as per policy.",
-                                              storageMetadata.getStorageLocation()));
+                // The metadata differs, but for valid reasons (update).
```
pdowler, line 108: In either case below (storageMetadata.isValid() true|false) the action is the same: 
```
validateEventListener.delete(storageMetadata);
validateEventListener.markAsNew(artifact);
```
so that a new copy of the file can be synced. It is worth reporting the invalid stored file so that if that happens we'll be able to see evidence of it and try to figure out the cause (in principle, it should "never" happen, but it will...)

Aside: "corrupt" may be too strong a term that makes operators panic :-) "incomplete" or "missing" is really what happens (eg in the file system extended file attributes are used to store the contentChecksum... the most likely case is a filed without that attribute rather than the attribute is there but the value was corrupted. Not saying that corruption is not possible, but I'd not use that term unless we decide to put an error message  into an invalid StorageMetadata and that message says "corrupt" because the adapter can detect that it did set attrs but one or more were actually corrupted.

tantar/src/main/java/org/opencadc/tantar/policy/InventoryIsAlwaysRight.java
```diff
@@ -95,25 +95,31 @@ public void resolve(final Artifact artifact, final StorageMetadata storageMetada
...
+                if (storageMetadata.isValid()) {
```
pdowler, line 109: I feel like the isValid check should be in the outer sequence of if/else and not inside the haveDifferentStructure because it is more fundamental (not valid is not equivalent to storageMetadata == null but closer to that than just "mismatched metadata")... basically, if not valid, the metadata comparison in haveDifferentStructure should not even be performed.

Also, if storageMetadata.isValid() == false, you do need to both `delete(storageMetadata)` and `markAsNew(artifact)`

Aside; use of the word "corrupt" here is probably overstating things most of the time and is likely to make ops people panic... they will interpret "files are being corrupted" to mean something else

at88mph: Thanks.  I've added some rework for messages and the `isValid` check.  I've modified the message to remove `Corrupt` for the `StorageIsAlwaysRight` policy, but corruption _may_ actually be the cause.

PR #161, opened 2020-08-10
s2767 - added allowRead & allowWrite to StorageSite in tap_schema

jburke-cadc:


luskan/src/main/resources/sql/inventory.tap_schema_content11.sql
```diff
@@ -57,7 +57,9 @@ insert into tap_schema.columns11 (table_name,column_name,utype,description,unit,
 ( 'inventory.StorageSite', 'name', 'si:StorageSite.name', 'Name of the storage site', NULL, 'char','32*',NULL, 1, 0, 1, 2 ),
 ( 'inventory.StorageSite', 'id', 'si:StorageSite.id', 'primary key', NULL, 'char','36','uuid', 1, 1, 1, 3 ),
 ( 'inventory.StorageSite', 'lastModified', 'si:StorageSite.lastModified', 'timestamp of the file at location', NULL, 'char','23*','timestamp', 1, 1, 1, 4 ),
-( 'inventory.StorageSite', 'metaChecksum', 'si:StorageSite.metaChecksum', 'checksum of the file at location', NULL, 'char','136*','uri', 1, 0, 1, 5 )
+( 'inventory.StorageSite', 'metaChecksum', 'si:StorageSite.metaChecksum', 'checksum of the file at location', NULL, 'char','136*','uri', 1, 0, 1, 5 ),
+( 'inventory.StorageSite', 'allowRead', 'si:StorageSite.allowRead', 'storage site allows HEAD and GET', NULL, 'boolean',NULL,'boolean', 1, 0, 1, 6 ),
```
pdowler, line 61: Don't need xtype=boolean for normal types -- should be null. xtype says that the raw value as described by datatype,arraysize can be interpretted as something more specialised but that doesn't apply here.

PR #163, opened 2020-08-12
S2767 - raven PUT support, move permission checking to cadc-inventory-util class

jburke-cadc:


raven/src/intTest/java/org/opencadc/raven/NegotiationTest.java
```diff
@@ -180,12 +176,81 @@ public Object run() throws Exception {
...
+                        siteDAO.put(site1);
+                        siteDAO.put(site2);
+
+                        SiteLocation location1 = new SiteLocation(site1.getID());
+                        SiteLocation location2 = new SiteLocation(site2.getID());
+
+                        Protocol protocol = new Protocol(VOS.PROTOCOL_HTTPS_PUT);
+                        Transfer transfer = new Transfer(
+                            artifactURI, Direction.pushToVoSpace, Arrays.asList(protocol));
+                        transfer.version = VOS.VOSPACE_21;
```
pdowler, line 215: To be complete, the transfer negotiation test should ask for multiple protocols:
https anon (no securityMethod -- the one that's there)
https w/ cert
https w/ cookie
https w/ token
http anon

The correct response would include:
1x protocol(https URL with pre-auth, securityMethod null) aka anon per site
3x protocol(https URl to minoc, securityMethod = {credential type}) per site
no http protocol (because minoc doesn't support http)

In principle, the protocols w/ securityMethod may or may not include the pre-auth stuff. TBD  so don't change whatever is there. All of this applies to GET and PUT.

pdowler (changes requested):
The PermissionCheck clas should really go in the cadc-inventory-server library since it is a server side component and will bring in dependencies that clients don't need. Use the org.opencadc.inventory.server.auth package 

The femwick test changes are also in PR #168 along with more substantive changes so I suggest reverting those and waiting for the other PR since this work should only involve raven and minoc (refactoring).

Comment about NegotiationTest below may cause changes so I have not looked at PostAction yet.

pdowler:
transfer to different developer

PR #164, opened 2020-08-26
Fixes #148

### LINKED ISSUE #148: critwall: not running in correct security context with Subject.doAs

at88mph:
Also ensure that exceptions are passed up.

### DIFF
--- a/critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
+++ b/critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
@@ -68,6 +68,7 @@
 package org.opencadc.critwall;
 
 import ca.nrc.cadc.auth.AuthMethod;
+import ca.nrc.cadc.auth.AuthenticationUtil;
 import ca.nrc.cadc.io.WriteException;
 import ca.nrc.cadc.net.FileContent;
 import ca.nrc.cadc.net.HttpGet;
@@ -90,6 +91,10 @@ import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
+import java.security.AccessControlContext;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
@@ -103,6 +108,9 @@ import org.opencadc.inventory.storage.StorageAdapter;
 import org.opencadc.inventory.storage.StorageEngageException;
 import org.opencadc.inventory.storage.StorageMetadata;
 
+import javax.security.auth.Subject;
+
+
 public class FileSyncJob implements Runnable {
     private static final Logger log = Logger.getLogger(FileSyncJob.class);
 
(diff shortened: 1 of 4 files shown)


critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -126,16 +131,32 @@ public FileSyncJob(URI artifactID, URI locatorServiceID, StorageAdapter storageA
...
+
     @Override
     public void run() {
         log.info("fileSyncJob start.");
+        if (this.owner == null) {
+            doSync();
+        } else {
+            Subject.doAs(this.owner, (PrivilegedAction<Void>) () -> {
+                doSync();
+                return null;
```
pdowler, line 146: Should not have a `return null` here. Surprised compiler didn't complain...

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -126,16 +131,32 @@ public FileSyncJob(URI artifactID, URI locatorServiceID, StorageAdapter storageA
...
+        log.info("doSync start.");
         final List<URL> urlList;
         try {
-            urlList = getdownloadURLs(this.locatorService, this.artifactID);
+            urlList = getDownloadURLs(this.locatorService, this.artifactID);
         } catch (Exception e) {
             // fail - nothing can be done without the list, negotiation cannot be retried
             log.error("transfer negotiation failed.");
-            return;
+            throw new RuntimeException(e);
```
pdowler, line 159: Can you please put the causing exception in the log.error call? When this failure happens all detailed info is lost.

critwall/src/main/java/org/opencadc/critwall/Main.java
```diff
@@ -229,8 +237,31 @@ public static void main(String[] args) {
                 "org.postgresql.Driver",
                 dbUrl);
 
-            FileSync doit = new FileSync(daoConfig, cc, localStorage, locatorService, bucketSel, nthreads);
-            doit.run();
+            final Subject subject = SSLUtil.createSubject(new File(CERTIFICATE_FILE_LOCATION));
```
pdowler, line 240: Two things here:
1. The application should work without a certificate so this needs to check if the file exists before creating a Subject.
2. The FileSync ctor should not do any remote calls that require auth, so only the `run()` needs to be inside the Subject.doAs. It's bad practice to have setup (ctors) require auth so we want to find and fix that kind of thing.
3. There is a standard wrapper (RunnableAction) you can use to execute a Runnable inside a Subject.doAs

pdowler (changes requested):

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
at88mph: The `PrivilegedAction` needs to return something though.  Is there a different way to do it?

at88mph: Actually, changed to be a `RunnableAction` instead.

pdowler (approved):

PR #175, opened 2020-09-09
Ad storage query support invalid metadata

jburke-cadc:


pdowler (approved):
I'd usually complain about `log.info` in a library, but this is temporary and only used in tantar internally.

PR #177, opened 2020-09-09
s2768 - updated luskan query with is not null sitelocations constraint

jburke-cadc:


luskan/src/main/java/org/opencadc/luskan/SiteLocationsConverter.java
```diff
@@ -0,0 +1,118 @@
...
+        } else {
+            Parenthesis parenthesis = new Parenthesis(where);
+            Expression and = new AndExpression(parenthesis, isNotNullExpression);
+            ps.setWhere(and);
+        }
+    }
+
+    private Expression isNotNullExpression() {
+        Table artifact = new Table("inventory", "artifact");
+        Column sitelocations = new Column(artifact, "sitelocations");
```
pdowler, line 111: This is not the correct column to constrain. Look at the model and documentation to see which things are used at storage sites vs global inventory.

luskan/src/test/java/org/opencadc/luskan/SiteLocationsConverterTest.java
```diff
@@ -0,0 +1,171 @@
...
+public class SiteLocationsConverterTest {
+    private static final Logger log = Logger.getLogger(SiteLocationsConverterTest.class);
+
+    static {
+        Log4jInit.setLevel("org.opencadc.luskan", Level.INFO);
+    }
+
+    private static final TapSchema tapSchema = TestUtil.loadTapSchema();
+
+    private static final String selectList =
```
pdowler, line 93: Having this selectList as an expected expansion of select * is testing some other code in the tap libs... and it's likely to be brittle because there is no correct order for the expansion.

luskan/src/test/java/org/opencadc/luskan/SiteLocationsConverterTest.java
```diff
@@ -0,0 +1,171 @@
...
+        "inventory.artifact.uri, inventory.artifact.uribucket, inventory.artifact.contentchecksum, "
+            + "inventory.artifact.contentLastmodified, inventory.artifact.contentlength, "
+            + "inventory.artifact.contenttype, inventory.artifact.contentencoding, inventory.artifact.lastmodified, "
+            + "inventory.artifact.metachecksum, inventory.artifact.id ";
+    private static final String fromClause = "FROM inventory.artifact";
+    private static final String isNotNullConstraint = "(inventory.artifact.sitelocations IS NOT NULL)";
+
```
pdowler, line 100: The relevant variants to test here are:
1. vanilla: FROM inventory.Artifact --> ({column} IS NOT NULL}
2. table alias: FROM  inventory.Artifact AS a --> (a.{column} IS NOT NULL)

So you have to find the table reference in the from clause and figure out whether or not to use the alias.

3. join: FROM inventory.Artifact a join inventory.Artifact b on a.id=b.id

Ok, it's hard to find a useful thing to do with that, but it should add a constraint for each alias)

luskan/src/test/java/org/opencadc/luskan/SiteLocationsConverterTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @Test
+    public void testSelectColumnWithAlias() {
+        String query = "select uri as \"foo\" from inventory.artifact";
+        String expected = "SELECT uri as \"foo\" " + fromClause + " WHERE " + isNotNullConstraint;
+        doTest(query, expected, "true");
+    }
+
+    @Test
+    public void testSelectWithWhere() {
+        String query = "select uri from inventory.artifact where uribucket = 1";
```
pdowler, line 124: this test is fine except that uriBucket is a char column so the value should have single ticks. It doesn't matter for the purposes of query mangling but it's distracting... maybe `contentLength <= 1024` would be less distracting.

luskan/src/test/java/org/opencadc/luskan/SiteLocationsConverterTest.java
```diff
@@ -0,0 +1,171 @@
...
+            List<Parameter> params = new ArrayList<Parameter>();
+            params.add(new Parameter("QUERY", query));
+            log.debug("query: " + query);
+            TapQuery tq = new TestQuery(isStorageSite);
+            tq.setTapSchema(tapSchema);
+            TestUtil.job.getParameterList().addAll(params);
+            tq.setJob(TestUtil.job);
+            log.debug("expected: " + expected);
+            String sql = tq.getSQL();
+
```
pdowler, line 141: I would add a `log.info("actual: " + sql)` and probably change the previous (log query and expected) to info as well.

pdowler (changes requested):
If two pieces of code need to access the luskan config, maybe add a LuskanConfig class; the config has nothing to do with initialising the database tables and content so does not belong in InitLuskanSchemaContent.

New configuration item *must* be added to the README.

luskan/src/main/java/org/opencadc/luskan/SiteLocationsConverter.java
jburke-cadc: Fixed column name.

luskan/src/test/java/org/opencadc/luskan/SiteLocationsConverterTest.java
jburke-cadc: Removed the expanded selectList from the tests.

luskan/src/test/java/org/opencadc/luskan/SiteLocationsConverterTest.java
jburke-cadc: Updated unit tests with the above cases.

luskan/src/test/java/org/opencadc/luskan/SiteLocationsConverterTest.java
jburke-cadc: Updated the test query as above.

luskan/src/test/java/org/opencadc/luskan/SiteLocationsConverterTest.java
jburke-cadc: Updated the logging as above.

luskan/src/main/java/org/opencadc/luskan/AdqlQueryImpl.java
```diff
@@ -128,4 +135,8 @@ public String getSQL() {
         log.debug("SQL:\n" + sql);
         return sql;
     }
+
+    protected MultiValuedProperties getProperties() {
```
pdowler, line 139: One line method - would be more clear to just call the static method above.

luskan/src/test/java/org/opencadc/luskan/StorageLocationConverterTest.java
```diff
@@ -0,0 +1,194 @@
...
+    private static final Logger log = Logger.getLogger(StorageLocationConverterTest.class);
+
+    static {
+        Log4jInit.setLevel("org.opencadc.luskan", Level.INFO);
+    }
+
+    private static final TapSchema tapSchema = TestUtil.loadTapSchema();
+
+    @Test
+    public void testSchemaName() {
```
pdowler, line 95: These queries should not have the `11` on the table name.

luskan/src/test/java/org/opencadc/luskan/TestUtil.java
```diff
@@ -0,0 +1,239 @@
...
+import ca.nrc.cadc.tap.schema.TapDataType;
+import ca.nrc.cadc.tap.schema.TapSchema;
+import ca.nrc.cadc.uws.Job;
+
+public class TestUtil {
+
+    /**
+     * load a TAP Schema for test purpose.
+     *
+     */
```
pdowler, line 86: Why two methods when there is really one? 

Also, this is going to eventually diverge from reality and be misleading, but I see it is necessary to write unit tests, so a couple of ways to minimise that:
1. don't need any tap_schema self description - you have tables where queries don't get mangled in the inventory schema - or at least strip it down to just the tap_schema.schemas table and use that
2. you could get by with a minimal tap_schema description of inventory tables: 4 tables with one column each (id column) would be enough... then it would be more clear than this isn't supposed to be the correct description

basically, it will be better if this has less content

luskan/src/test/java/org/opencadc/luskan/StorageLocationConverterTest.java
```diff
@@ -0,0 +1,194 @@
...
+    public void testSelectWithTableAlias() {
+        String query = "select uri from inventory.artifact a";
+        String expected = "SELECT uri from inventory.artifact AS a WHERE (a.storagelocation_storageid IS NOT NULL)";
+        doTest(query, expected, "true");
+    }
+
+    @Test
+    public void testSelectWithTableJoin() {
+        String query = "select a.uri from inventory.artifact a join inventory.artifact b on a.id = b.id";
+        String expected = "SELECT a.uri from inventory.artifact AS a JOIN inventory.artifact AS b on a.id = b.id WHERE (a.storagelocation_storageid IS NOT NULL)";
```
pdowler, line 139: The expected query is wrong here - there should also be a 
```
(b.storagelocation_storageid IS NOT NULL)
```
in there. See comment in StorageLocationCnverter

luskan/src/main/java/org/opencadc/luskan/StorageLocationConverter.java
```diff
@@ -0,0 +1,131 @@
...
+ */
+public class StorageLocationConverter extends SelectNavigator {
+    private static final Logger log = Logger.getLogger(StorageLocationConverter.class);
+
+    public StorageLocationConverter() {
+        super(new ExpressionNavigator(), new ReferenceNavigator(), new FromItemNavigator());
+    }
+
+    @Override
+    public void visit(PlainSelect ps) {
```
pdowler, line 97: you must call `super.visit(ps)` first to recurse correctly. Yeah, no test have subqueries (which is fine for now).

If you look at CaomReadAccessConverter, you will see use of `ParserUtil.getFromTableList(ps)` followed by a loop over all the tables in the from clause. Probably could split up the more complex code like it is in CaomReadAccessConverter.visit (which is pretty easy to understand by itself)... then put the complex part in a method like `createConstraint(PlainSelect)` that returns a `null` or `Parenthesis`

pdowler (changes requested):
There should be no change to InitLuskanSchemaContent - still some left over changes in there.

luskan/src/main/java/org/opencadc/luskan/AdqlQueryImpl.java
jburke-cadc: Made the protected one liner to allow unit test to have different luskan.properties configurations. Can toggle between a config for a storage site that should have the extra constraint injected in the query, and a global that should not.

luskan/src/test/java/org/opencadc/luskan/StorageLocationConverterTest.java
jburke-cadc: Removed the tap_schema tables from the mock tap schema.

luskan/src/test/java/org/opencadc/luskan/TestUtil.java
jburke-cadc: Removed the tap_schema tables and used a minimal inventory description.

luskan/src/test/java/org/opencadc/luskan/StorageLocationConverterTest.java
jburke-cadc: Updated the unit test with the extra constraint.

luskan/src/main/java/org/opencadc/luskan/StorageLocationConverter.java
jburke-cadc: The first time tracing through visit(PlainSelect) I thought I saw super.visit(ps) had already been called in the SelectNavigator, but don't see it now.
Updated to account for join tables.

luskan/src/main/java/org/opencadc/luskan/AdqlQueryImpl.java
```diff
@@ -115,6 +116,12 @@ protected void init() {
         TableNameReferenceConverter tnrc = new TableNameReferenceConverter(tnc.map);
         super.navigatorList.add(new SelectNavigator(new ExpressionNavigator(), tnrc, tnc));
 
+        // add IS NOT NULL constraint for artifact.storagelocation_storageid when querying storage sites
+        MultiValuedProperties properties = getProperties();
+        String isStorageSite = properties.getFirstPropertyValue(LuskanConfig.STORAGE_SITE_KEY);
+        if (isStorageSite != null) {
```
pdowler, line 122: Since the README says
```
org.opencadc.luskan.isStorageSite={true|false}
```
Shouldn't this be something like
```
boolean isStorageSite = Boolean.parseBoolean(...);
if (isStorageSite) {
...
```
As it is, `false` and 'flibble` will also make `if (isStorageSite != null) {` true.

luskan/src/main/java/org/opencadc/luskan/LuskanConfig.java
```diff
@@ -0,0 +1,123 @@
...
+
+        String srd = props.getFirstPropertyValue(TMPDIR_KEY);
+        sb.append("\n\t").append(TMPDIR_KEY);
+        if (srd == null) {
+            sb.append("MISSING");
+            ok = false;
+        } else {
+            sb.append("OK");
+        }
+
```
pdowler, line 116: README implies that STORAGE_SITE_KEY is mandatory - check for it here and add to failure message. Also validate the value with Boolean.parseBoolean

pdowler (changes requested):

luskan/src/main/java/org/opencadc/luskan/LuskanConfig.java
jburke-cadc: If it's mandatory that should be validated here. But I'm not sure these checks are doing what is intended. Looking at the MultiValueProperties class, **getFirstPropertyValue(key)** will return null if the property doesn't exist. These checks are verifying that the properties have a value, not that they are set.

luskan/src/main/java/org/opencadc/luskan/AdqlQueryImpl.java
jburke-cadc: Wasn't sure if this property should be mandatory at first so did the null check. Forgot to revisit after decided it should be, but parseBoolean is the proper way to get the property value.

luskan/src/main/java/org/opencadc/luskan/LuskanConfig.java
jburke-cadc: For properties that have a string value, like resourceID, a null value is an error. But for a boolean, a null value defaults to false.

pdowler: It is mandatory. This is verifying that the config file is complete and throws an exception with a useful message if not.

pdowler (approved):

PR #187, opened 2020-09-18
s2768 - tantar: log a null uri in the resultset and move on to the next row.

jburke-cadc:


pdowler (changes requested):
libraries NEVER log at info level

Why is AdStorageMetadataRowMapper keeping a reference to the StorageMetadata it returns? RowMapper's never do that - they just converta  row into an object. It looks like it will return the same row twice if the row being mapped has no URI. That's just wrong.

The AdStorageAdapter already had  away to skip items (duplicates) so you might want to find out how that works and also skip nulls.

jburke-cadc:
Changed the logging level to debug.

I wrote another unit test with a null StorageMetadata object, and stepped through the AdStorageIterator a couple of times, finally noticing the magic is in the while loop in next(). If the AdStorageMetadataRowMapper returns a null when it can't create a StorageMetadata, the AdStorageIterator does skip that row.

The exception is if the first StorageMetadata object returned by the AdStorageMetadataRowMapper is null. In that case the AdStorageIterator exits, because hasNext() returns false given the first iterator object is null.

The AdStorageIterator constructor could be updated to check if the first row(s) are null.
```
while (this.nextItem == null && sourceIterator.hasNext()) {
            this.nextItem = sourceIterator.next();
        }
```
Or it could be a known issue. A small amount of data (~600 rows of CFHT data) out of ~30 million rows in total can't create a StorageMetadata object.

pdowler (changes requested):
Yes: the  AdStorageIterator was written to skip duplicates and the ctor was correct when the underlying iterator never returned null; not that is no longer true so it has to be fixed to skip nulls until it finds the first item. I've often implemented a method like `private void advance()` but it's tricky here because of skipping nulls and duplicates.

jburke-cadc:
Updated the constructor to loop until the first non-null object is returned from the sourceIterator.

pdowler (approved):

Issue #195, opened 2020-09-26
fenwick: not necessary to harvest all types of events to every destination


pdowler:
storage site -> global (trackSiteLocations == true), fenwick should harvest:
- StorageSite
- DeletedStorageLocationEvent
- DeletedArtifactEvent
- Artifact

global -> storage site (trackSiteLocations == false), fenwick should harvest:
- DeletedArtifactEvent
- Artifact

pdowler:
secondary improvement in fenwick: 

InventoryHarvester.syncDeletedStorageLocationEvents does not need to persist the events to the local database because unlike DeletedArtifactEvent(s) they do not have to propagate from global to other storage sites. These events are strictly for global to discover that a site lost it's copy of a file.

pdowler:
Secondary improvement: DeletedStorageLocationEvent do not need to be stored in the destination database because they don't propagate from global (unlike DeletedArtifactEvent which does propagate). I missed that in earlier code reviews.

PR #199, opened 2020-09-30
Fixes 198

at88mph:


pdowler (changes requested):
This is not really the correct fix. See issue #195 which would avoid this NPE and a bunch of useless work by fenwick.

at88mph:
No longer relevant.

Issue #166, opened 2020-09-02
fenwick: multiple included condition files was a bad idea


pdowler:
Each include file defines a separate stream of new|modified Artiact events; that is hard to manage and won't play well with also processing DeletedArtifactEvent(s) in a timely fashion.

Should just be one optional /config/artifact-filter.sql (ish) file

PR #205, opened 2020-10-07
WIP: Fixes #201

### LINKED ISSUE #201: fenwick: long term execution
Currently, fenwick does all init up front and then enters an infinite loop. This includes: database connections and client certificate for use in transfer negotiation and download. Connections can go stale or get dropped and the certificate in memory could expire while the file on disk gets updated.

Changes:
* move database connection setup inside loop, before the queries; cleanup and close connections when jobs are done and before sleep
* move certificate->subject into the loop, maybe re-loading the Subject before each query

at88mph:
* Added pre-check for `artifact-filter.sql` include file.  Bug from a fix for #166.
 * Moved Subject creation to before each sync job.

Looking for acceptance for this part while carrying on with moving the DataSource allocation.

### DIFF
--- a/fenwick/src/main/java/org/opencadc/fenwick/IncludeArtifacts.java
+++ b/fenwick/src/main/java/org/opencadc/fenwick/IncludeArtifacts.java
@@ -91,7 +91,7 @@ import org.apache.log4j.Logger;
 public class IncludeArtifacts implements ArtifactSelector {
     private static final Logger log = Logger.getLogger(IncludeArtifacts.class);
     static final String COMMENT_PREFIX = "--";
-    static final String SQL_FILE_NAME = "artifact-filter.sql";
+    static final String SQL_FILTER_FILE_NAME = "artifact-filter.sql";
 
     private final Path selectorConfigDir;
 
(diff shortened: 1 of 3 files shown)


pdowler (commented):
I was thinking about having a single Subject.doAs(...) and then periodically reading the cert and updating it (in the current Subject). That way the program flow/logic doesn't change and the details of keeping the cert up-to-date are separate/orthogonal (easiest method: a method that checks if the current cert is near expiry and re-reads the file -- then call that method somewhere in the loops). Such a method could be tweaked to only check occasionally or removed from the loop and put on a timer.

On that topic, please revert the "skip first entity" logic to the way it was. Yes, `continue` is kind of ugh but the logic and it's effect is much more clear than not and or and the effect being way down in the else.

at88mph:
Thanks.  I've updated again.  Updating the current Subject now involves replacing the Principals as well as the Credentials with the use case being a different user's certificate will be used.

pdowler (approved):

pdowler:
This did not address the issue of renewing database connections, but I realise that is better addresses by making the JNDI DataSource into a connection pool that recycles connections anyway (not that fewick needs it, but critwall does need it so it will be implemented at some point).

Issue #185, opened 2020-09-16
cadc-storage-adapter-ad: Unauthorized calls not returning properly.


at88mph:
The `AdStorageAdapter` is not handling `AccessControlException`s properly, or is failing to propagating them.  The call to the `data` web service could result in a 403 or a 401, which should be sent along to the client.  Currently it returns a successful result code, with an `Unauthorized` message.

pdowler:
AdStorageAdapter needs to use delegated credentials correctly when proxying the call to `data`

PR #207, opened 2020-10-08
S2803 proxy call to data #185

jburke-cadc:


cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageAdapter.java
```diff
@@ -144,6 +147,14 @@ public void get(StorageLocation storageLocation, OutputStream dest)
         URL sourceURL = this.toURL(storageLocation.getStorageID());
         log.debug("sourceURL: " + sourceURL.toString());
 
+        try {
+            boolean auth = CredUtil.checkCredentials();
+            log.debug("authenticated: " + auth);
+        } catch (CertificateExpiredException | CertificateNotYetValidException unexpected) {
+            log.debug("error type: " + unexpected.getClass());
```
pdowler, line 154: This has to be an AccessControlException with a clear reason... something like 

```
permission denied -- reason: delegated X509 Certificate is invalid
```

Also: it is definitely *not* unexpected and the debug message can be removed

pdowler (changes requested):

cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageAdapter.java
jburke-cadc: Changed exception type and message, and removed the debug message.

pdowler (approved):

Issue #191, opened 2020-09-23
luskan: start up successul even though config is invalid


pdowler:
at a minimum: error messages in the log on startup and /availability reporting invalid config

Issue #81, opened 2020-03-05
luskan: AvailabilityPlugin does not check that TempStorageManager works


pdowler:
should check for config and check that the location is writable

PR #208, opened 2020-10-08
S2803 luskan startup issues #81

jburke-cadc:
Issue #191 was fixed in the last pull request in the luskan query correctness work.
Added a check in the availability test that writes a temp file in the results directory.
The TempStorageManager constructor does a fairly complete job of checking that properties and directories are in correct. The check to ensure the results directory is writable could be added to the constructor to log the error when the service starts. If that was done, would the check be needed in availability?


luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
```diff
@@ -252,4 +254,17 @@ private String getState() {
         return ret;
     }
 
+    // Check that the TempStorageManager results directory is writable.
```
pdowler, line 257: If you were to put this method into LuskanConfig and call it from initConfig(), then the check would also run on startup and throw an exception -- not just in the availability.

luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
```diff
@@ -252,4 +254,17 @@ private String getState() {
...
+    private void checkTempStorage() {
+        String tmpDir = LuskanConfig.getConfig().getFirstPropertyValue(LuskanConfig.TMPDIR_KEY);
+        Path tmpDirPath = Paths.get(tmpDir);
+        try {
+            Path tmpFile = Files.createTempFile(tmpDirPath, null,null);
+            Files.delete(tmpFile);
+        } catch (IOException e) {
+            throw new RuntimeException(String.format(
+                "Unable to write to TempStorageManager directory %s", tmpDirPath.toAbsolutePath()));
```
pdowler, line 266: In the filesystem storage adapter, the check for read-write permission fails with an IllegalArgumentException -- that is the general exception for bad input. RuntimeException should only be used for for a more extreme case like detected bug.

pdowler (changes requested):

luskan/src/main/java/org/opencadc/luskan/ws/StorageTapService.java
jburke-cadc: That's a better place for that check, moved to LuskanConfig.

pdowler (approved):

PR #211, opened 2020-10-09
Fixes #182

### LINKED ISSUE #182: critwall: long term execution
Currently, critwall does all init up front and then enters an infinite loop. This includes: database connections and client certificate for use in transfer negotiation and download. Connections can go stale or get dropped and the certificate in memory could expire while the file on disk gets updated.

Changes:
- move certificate loading into the job creation loop, maybe re-loading the certificate after some fixed amount of time rather than once per loop or a number of jobs? maybe jusr pass cert file name into FileSyncJob?

at88mph:
* Added scheduler to refresh Subject
 * Moved Subject tasks into FileSync
 * Removed some string builder warnings


pdowler (changes requested):
Every 5 hours the scheduled task checks if the current cert has less than 5 hours left and re-reads the file. So if there is 5:00:01 left then next time there will be 0:00:01 left. Reading the file is pretty cheap, so I would schedule more frequent checks (like every 10-30 minutes or so).

The fenwick usage was so simple so I did not appreciate the following: I don't think impending expiry is the only reason to switch to a new cert: if ops puts a new cert in place it should get picked up and used in a timely fashion. This wasn't so obvious in fenwick because the cert was checked right before use. 

**Change 1**: please change to ~10 minute timer and just load the new cert into the current subject every time. Since you have to read the file to do a check if it is different (principal or expiry) anyway, might as well just use that cert even if it is the same.

**Change 2**: the `verifySubject` method could be re-usable by fenwick if it was in a library, say as `public static renewSubject(Subject subject, File certFile)` ... could go into cadc-inventory-util (AuthUtil?) or even all the way to cadc-util (SSLUtil).

The real problem: 

In this section of code (called by scheduled thread), currently lines 214-218:
```
currentSubject.getPrincipals().clear();
currentSubject.getPrincipals().addAll(subject.getPrincipals());

currentSubject.getPublicCredentials().clear();
currentSubject.getPublicCredentials().addAll(subject.getPublicCredentials());
```
the Subject is in an inconsistent state and possibly lacks credentials, so FileSyncJob(s) running in the thread pool will fail if they happen to access the subject at the wrong time: classic race condition where the outcome cannot be predicted. So this needs to synchronize access to the Subject. The hard part is that the Subject is used in this verify method and in HttpTransfer so there is no good/obvious way to synchronize the actual access. But this should work and be robust:

**Change 3**:
1. put the subject-altering lines in a `synchronized(currentSubject)` block (javadoc should mention synchronized on the subject)
2. change the run() method in FileSyncJob to safely make a local copy of the subject, eg
```
Subject currentSubject = new Subject();
synchronized (subject) {
    currentSubject.getPrincipals().addAll(subject.getPrincipals());
    currentSubject.getPublicCredentials().addAll(subject.getPublicCredentials());
}
Subject.doAs(currentSubject, new RunnableAction(this::doSync));
```

Can also change fenwick to use the SSLUtil.renewSubject method.

at88mph:
Having the synchronized access in Critwall makes sense; the scheduler runs every ten minutes, and the `FileSyncJob` instances each have a `Subject` in them so they synchronize read access to them first.

Fenwick, however, creates the `Subject` once and the entire `doit` is performed, which means it would need to synchronize on the `AccessController`'s current `Subject`.  Should Fenwick just keep those three calls it has now to update in each of the loops and not use the scheduler?  Maybe the `InventoryHarvester` could have its own `Subject` and it gets update the same way as it does in the `FileSyncJob`?

at88mph:
Let me clarify actually; Fenwick is single threaded and so shouldn't need any synchronization as it won't be running on a scheduler.  It would be nice if it did, but I think it would require modifying the way the `InventoryHarvester` is setup.

Critwall's `FileSyncJob`s, however, do need synchronization as the `currentSubject` is passed into it and could be updated at a bad time.  Modifying the `Subject` (i.e. running `SSLUtil.renewSubject`) should also be synchronized, but without a way to lock down access to the `AccessController`'s current subject, I'm not sure that can be done.  Alternatively, the `FileSync` could store the `Subject` as a member and that's the one that gets synchronized on.

pdowler:
In both applications, there is only ever one instance of Subject. 

for the library function, `synchronized(subject)` is fine even if no one else is also doing it. for fenwick, because it is single threaded and renews subject right before query it should be find using the library method and nothing else. 

for critwall, all the FileSyncJob instances have a ref to the singleton Subject instance, so `synchronized(subject)` will work as intended. By carefully making a local copy of the subject and using that in their own Subject.doAs, FileSyncJob would be safe from subsequent unlucky changes in the "outer" subject.

Alternatively, since the lib method is renewSubject(Subject s, File cert), the  caller could be responsible for synchronized; it's a courser grained sync block but might be more clear and better lib design.  In this alternative, FileSync would have to also do `synchronized(subject)` around the call to renewSubject because it is running in the main thread. 

Having the `synchronized` blocks in FileSync and FileSyncJob means that a comment right there saying which other  `synchronized(subject)` this is competing with refers to the same piece of s/w. That's going to make more sense in the future because whenever you see  `synchronized` you want to know where the other one is, so I like to see something like this in FileSync:
```
// synchronized vs FileSyncJob.run() 
synchronized(subject) {
...
}
```
and the opposite comment in FileSyncJob.run().

at88mph:
Great, thank you.  That is all good as I like having the `SSLUtil.renewSubject` clean and callers will be responsible for synchronizing the passed in Subject.  The part that doesn't make sense still is how the scheduler in Critwall will wrap the `renewSubject()`.

Consider:

```FileSync.java
final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
scheduledExecutorService.schedule(this::renewSubject, CERT_CHECK_MINUTE_INTERVAL_COUNT,
                                          TimeUnit.MINUTES);
```

which could be changed to:

```FileSync.java
final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
final Subject currentSubject = AuthenticationUtil.getCurrentSubject();
scheduledExecutorService.schedule(() -> {
    synchronized(currentSubject) {
        SSLUtil.renewSubject(currentSubject, certFile);
    }
}, CERT_CHECK_MINUTE_INTERVAL_COUNT,
                                          TimeUnit.MINUTES);
```

But then the `currentSubject` is a local variable and being synchronized on which, I don't think, does anything unless it affects how it's accessed on the heap somehow.

at88mph:
I've updated this PR with some changes.  I think it will protect the Subject adequately.

cadc-inventory/src/main/java/org/opencadc/inventory/InventoryUtil.java
```diff
@@ -89,9 +95,36 @@
 public abstract class InventoryUtil {
     private static final Logger log = Logger.getLogger(InventoryUtil.class);
 
+    public static final String CERTIFICATE_FILE_LOCATION = System.getProperty("user.home") + "/.ssl/cadcproxy.pem";
+
```
pdowler, line 99: I can foresee use cases where some components have configured cert files and cannot operate with a single file in default location, so I'd prefer to leave this default in the individual components and not entrench this idea in the libraries with by defining constants and adding convenience methods.

So, components (critwall and fenwick currently) should call a `renewSubject(Subject, File)` method directly.

pdowler (changes requested):
One small change + cadc-util-1.3.12 is now published so CI should work.

at88mph:
Thank you.  Review rework committed.

pdowler (approved):

pdowler (approved):

Issue #56, opened 2020-02-05
luskan: async results are stored and never cleaned up


pdowler:
async query results are stored in a local directory (inside container by default) and never cleaned up.

PR #217, opened 2020-10-16
CADC-1309 luskan async query results cleanup issue #56 aka CADC-8532

jburke-cadc:
Tried to create a unit test for this, but too many preconditions. I did hack together a test to confirm it is deleting as expected after the time interval has elapsed.

### DIFF
--- a/luskan/src/main/java/org/opencadc/luskan/TempStorageManager.java
+++ b/luskan/src/main/java/org/opencadc/luskan/TempStorageManager.java
@@ -89,8 +89,11 @@ import java.net.URI;
 import java.net.URL;
 import java.sql.ResultSet;
 
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import org.apache.log4j.Logger;
-
+import org.opencadc.luskan.ws.QueryJobManager;
 
 /**
  * Basic temporary storage implementation for a tap service.
@@ -104,7 +107,7 @@ public class TempStorageManager implements ResultStore, UWSInlineContentHandler 
     private final File resultsDir;
     private String baseResultsURL;
 
-    //private Job job;
+    private Job job;
     private String filename;
 
     public TempStorageManager() {
@@ -158,6 +161,8 @@ public class TempStorageManager implements ResultStore, UWSInlineContentHandler 
         }
         // TODO: store requested content-type with file so that 
         // TempStorageGetAction can set content-type header correctly
+
+        scheduleDeletion(dest, getJobDestructionDelay());
         return ret;
     }
 
@@ -169,12 +174,14 @@ public class TempStorageManager implements ResultStore, UWSInlineContentHandler 
         }
         // TODO: store requested content-type with file so that 
         // TempStorageGetAction can set content-type header correctly
+
+        scheduleDeletion(dest, getJobDestructionDelay());
         return ret;
     }
 
     @Override
     public void setJob(Job job) {
-        // unused
+        this.job = job;
     }
 
     @Override
@@ -240,6 +247,8 @@ public class TempStorageManager implements ResultStore, UWSInlineContentHandler 
         fos.flush();
         fos.close();
 
+        scheduleDeletion(put, QueryJobManager.MAX_DESTRUCTION);
+
         URL retURL = new URL(baseResultsURL + "/" + filename);
         Content ret = new Content();
         ret.name = UWSInlineContentHandler.CONTENT_PARAM_REPLACE;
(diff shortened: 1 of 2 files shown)


luskan/src/main/java/org/opencadc/luskan/TempStorageManager.java
```diff
@@ -169,12 +174,14 @@ public URL put(Throwable t, TableWriter writer) throws IOException {
         }
         // TODO: store requested content-type with file so that 
         // TempStorageGetAction can set content-type header correctly
+
```
pdowler, line 177: Why can't this also be scheduled for deletion after QueryJobManager.MAX_DESTRUCTION like the inline content? I think the method results in approximately the same amount of time. The solution is simple so should be implemented as simply as possible.

luskan/src/main/java/org/opencadc/luskan/TempStorageManager.java
```diff
@@ -250,4 +259,35 @@ public Content accept(String name, String contentType, InputStream inputStream)
     private static String getRandomString() {
         return new RandomStringGenerator(16).getID();
     }
+
+    private long getJobDestructionDelay() {
+        return this.job.getDestructionTime().getTime() - System.currentTimeMillis();
+    }
+
+    void scheduleDeletion(final File file, final long delay) {
```
pdowler, line 267: Does this actually work? the method creates a new executor on every call and schedules one job in there... then it calls shutdown() on the executor. So either:
* the scheduled job never runs, or 
* the shutdown() call blocks for a long time, or
* that executor API is really misleading because this looks wrong

Anyway, even if it worked you don't want an idle thread sitting there for ~2 days for every async query and inline upload.

Since there is a ScheduledExecutorService with a schedule(...) method, this code really needs a singleton: that means create one and store it for re-use. Since instances of TempStorageManager get created by every request, you need a static or lazy initialiser that creates the ScheduledExecutorService and stores it for later use. I would put the lazy init in the scheduleDeletion method. Then the body of the method simply gets the stored executor and schedules another deletion... the executor can run forever (until tomcat shuts down). 

Store it could be as a JNDI resource or as a static variable in TempStorageManager. The tricky bit is that you have no local way to come up with a name for the JNDI resource that is unique to this one application (on the grounds that this class could be used in two different apps for some open source reason) but I'm not sure that matters - if the code named the resource "TempStorageManager/deletion-executor" then all apps would share an executor and get it to delete files for them... and if it is safe, then just having 
```
private static ScheduledExecutorService SCHEDULER = Executors.newSingleThreadScheduledExecutor();
```
would work the same way and also be fine. The only catch is whether or not the thing will shutdown cleanly or not (ie is it a normal or daemon thread?) whent he JVM tries to shutdown. Can you see any reason that's a bad idea?

pdowler (changes requested):

luskan/src/main/java/org/opencadc/luskan/TempStorageManager.java
jburke-cadc: Updated to use the QueryJobManager.MAX_DESTRUCTION for all deletions.

luskan/src/main/java/org/opencadc/luskan/TempStorageManager.java
jburke-cadc: Yes it does work. But yes it would also create X number of long lived threads.
Storing the scheduler service in the context would allow it to be retrieved and properly shutdown when the JVM terminates, but is the extra complexity worth it? A static shared among instances is simplest. Scheduled deletions not yet run could stop the jvm from exiting. From testing it does not stop the container from exiting. Using a ThreadFactory that returns daemon threads for the ScheduledExecutorService would allow the JVM to exit immediately. Deletion threads that happen to be running would be abruptly terminated, but for temp files I think that should be okay.

luskan/src/main/java/org/opencadc/luskan/TempStorageManager.java
```diff
@@ -250,4 +263,45 @@ public Content accept(String name, String contentType, InputStream inputStream)
...
+                        log.info(String.format("%s not found for deletion", file.getAbsolutePath()));
+                    } else {
+                        boolean deleted = file.delete();
+                        log.debug(String.format("deleted %s: %s", file.getAbsolutePath(), deleted));
+                    }
+                } catch (SecurityException e) {
+                    log.error(String.format("Unable to delete %s because %s", file.getAbsolutePath(),
+                                            e.getMessage()));
+                }
+            }
```
pdowler, line 281: The constants in QueryJobManager are in seconds.

pdowler (changes requested):

luskan/src/main/java/org/opencadc/luskan/TempStorageManager.java
jburke-cadc: That would be about 10 minutes before deletion. Fixed.

pdowler (approved):

PR #230, opened 2020-11-08
add VERSION file with semantic and build version tags; CADC-8612

pdowler:
simplified application docker files by removing version attr from build.gradle


brianmajor (approved):
The tagging in docker-base looks good.  I like the little script to help automate.

In storage-inventory, just wondering why the semantic and build tags are separate…

From the ops perspective, I suppose this approach offers flexibility.  If you reference the timestamp tag, releases must be explicitly made.  If you reference the semantic tag, you may get automated updates if the ‘pull policy’ is set to allow it.

Issue #216, opened 2020-10-15
tantar: does not consult ObsoleteStorageLocation list


pdowler:
When a file is deleted via minoc, the Artifact is deleted and an ObsoleteStorageLocation is created (in txn).
When a file is replaced via minoc, the old Artifact is deleted and *iff* the new StorageLocation is different (true for ceph and opaque-fs), an ObsoleteStorageLocation is created (in txn).

After the database txn is committed, the file is deleted from back end storage (StorageAdapter.delete). If that step succeeds, the ObsoleteStorageLocation is deleted. if it fails, the ObsoleteStorageLocation remains in the database so that correct cleanup can occur -- tantar.

Correct cleanup: when iterating through StorageMetadata instances from the StorageAdapter, tantar should check for a matching ObsoleteStorageLocation. If found, delete the file from storage and then delete the ObsoleteStorageLocation from the database. This should happen before consulting a policy to resolve Artifact vs StorageMetadata because it applies to all policies.

PR #235, opened 2020-11-18
ringhold implementation CADC-8572 

jburke-cadc:


ringhold/src/intTest/java/org/opencadc/ringhold/InventoryValidatorTest.java
```diff
@@ -0,0 +1,328 @@
...
+        this.artifactDAO.put(a2);
+        Artifact a3 = getTestArtifact("cadc:CADC/three.txt");
+        this.artifactDAO.put(a3);
+
+        InventoryValidator testSubject = new InventoryValidator(this.daoConfig, this.daoConfig);
+        testSubject.run();
+
+        ResourceIterator<Artifact> artifacts = this.artifactDAO.iterator("uri LIKE 'cadc:INTTEST/%'", null);
+        Assert.assertFalse(artifacts.hasNext());
+
```
pdowler, line 205: The above iterator and assert look like a re-run of what happens inside the InventoryHarvester... never like to see implementation code repeated in the tests. The checks below assert that the result is correct.

Comment applies to other test methods as well.

pdowler (changes requested):

ringhold/src/main/java/org/opencadc/ringhold/InventoryValidator.java
```diff
@@ -80,25 +86,71 @@
...
+     * delete the artifact and generate a deleted storage location event.
+     */
     @Override
     public void run() {
-        throw new UnsupportedOperationException("NOT IMPLEMENTED");
+        final TransactionManager transactionManager = this.artifactDAO.getTransactionManager();
+        final DeletedStorageLocationEventDAO deletedStorageLocationEventDAO =
+            new DeletedStorageLocationEventDAO(this.artifactDAO);
+
+        ResourceIterator<Artifact> artifactIterator = this.artifactIteratorDAO.iterator(this.deselector, null);
```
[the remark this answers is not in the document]
jburke-cadc: Used the try-with-resources and added the catch for the IOException to be more explicit where and why it's thrown, rather than declare run() throws an IOException and let the Main() try/catch would handle it.

ringhold/src/intTest/java/org/opencadc/ringhold/InventoryValidatorTest.java
```diff
@@ -0,0 +1,328 @@
...
+        DeletedStorageLocationEvent a_dsle1 = this.deletedStorageLocationEventDAO.get(a1.getID());
+        Assert.assertNotNull(a_dsle1);
+        DeletedStorageLocationEvent a_dsle2 = this.deletedStorageLocationEventDAO.get(a2.getID());
+        Assert.assertNotNull(a_dsle2);
+        DeletedStorageLocationEvent a_dsle3 = this.deletedStorageLocationEventDAO.get(a3.getID());
+        Assert.assertNotNull(a_dsle3);
+    }
+
+    private void writeConfig() throws IOException {
```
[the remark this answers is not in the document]
jburke-cadc: Makes sense. Put the artifact-deselector.sql in build/tmp/config and set and restore user.home before and after the run() method.

ringhold/src/intTest/java/org/opencadc/ringhold/InventoryValidatorTest.java
jburke-cadc: No it's not. It was checking if there are any results, not specific results, but the following checks do that too. Removed.

ringhold/src/main/java/org/opencadc/ringhold/InventoryValidator.java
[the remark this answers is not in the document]
pdowler: I think the only source of IOException is in the close() call: you don't want to throw the RuntimeException on line 157. If the close() fails, it is because something else went wrong and was probably caught up at line 139 or maybe from line 151... you want that exception to get thrown. A throw-overrides-throw is a common way to hide errors - this is the one scenario where you log an exception only.

BTW, why doesn't the throw on line 145 require run() to declare that it throws?

ringhold/src/intTest/java/org/opencadc/ringhold/InventoryValidatorTest.java
```diff
@@ -0,0 +1,323 @@
...
+        artifactDAO.setConfig(daoConfig);
+        deletedStorageLocationEventDAO.setConfig(daoConfig);
+    }
+
+    @Before
+    public void setup() throws Exception {
+        writeConfig();
+        truncateTables();
+    }
+
```
pdowler, line 181: Final thing I didn't notice the other day. I have generally found that it is best to do cleanup to a known state @before and not @after. That way, the last test run leaves it's content behind and if a test is failing a dev can run that one test and examine the database state manually afterward. Leaving a little content from the last test behind is generally fine - other tests using that database have to cleanup before to be safe anyway.

Also, in the constructor, you should invoke the InitDatabase facility in the db lib. Kill your pg container and start a new one to verify that works.

pdowler (changes requested):

ringhold/src/main/java/org/opencadc/ringhold/InventoryValidator.java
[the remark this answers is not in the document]
jburke-cadc: Updated to only catch and log the IOException for try-with-resources, removed the throws for a RuntimeException.
// BTW, why doesn't the throw on line 145 require run() to declare that it throws?
Good question, had to look that up. Short answer is no checked exceptions are thrown in the try that could be thrown again in the catch.

ringhold/src/intTest/java/org/opencadc/ringhold/InventoryValidatorTest.java
jburke-cadc: Updated the int test to use initDatabase, and that does work, just have to give the pg container a couple of secs to startup before running the intTest.

pdowler (dismissed):

ringhold/src/main/java/org/opencadc/ringhold/InventoryValidator.java
```diff
@@ -80,25 +86,74 @@
...
+            new DeletedStorageLocationEventDAO(this.artifactDAO);
+
+        try (final ResourceIterator<Artifact> artifactIterator =
+            this.artifactIteratorDAO.iterator(this.deselector, null)) {
+            while (artifactIterator.hasNext()) {
+                Artifact deselectorArtifact = artifactIterator.next();
+                log.debug("START: Process Artifact " + deselectorArtifact.getID() + " " + deselectorArtifact.getURI());
+
+                try {
+                    transactionManager.startTransaction();
```
pdowler, line 129: The bulk cleanup logic task (CADC-8663) specifies three things that have to happen in this transaction. Only two of them are here.

In addition, ringhold relies on an external process to clean up files. While we envisioned AD cleanup to be manual, in other sites that use ringhold to remove a set of artifacts it would be tantar. While looking at issues, #216 reminded me that there is a way to explicitly specify that a stored file is obsolete. So: rionghold will be more generally useful if it also creates an ObsoleteStorageLocation and persists it. This is a mechanism outside the model but supported in the db lib (see minoc DeleteAction) to be added to this txn (create and store an OSL).

ringhold/src/intTest/java/org/opencadc/ringhold/InventoryValidatorTest.java
```diff
@@ -0,0 +1,328 @@
...
+import org.opencadc.inventory.db.version.InitDatabase;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+/**
+ * Various versions of:
+ * Insert artifacts more than uri pattern
+ * Run tool with one uri deselector
+ * Confirm delete storage location event creation and absence of artifacts in inventory
+ */
+public class InventoryValidatorTest {
```
pdowler, line 112: There should be a test with a faulty (dangerous) config that verifies that this will not remove a bunch of content when misconfigured. I'm not concerned with the test verifying that the ERROR log messages get printed - that can be seen in code review - but the test should verify that nothing gets deleted in misconfigured cases:

- missing deselector file
- empty deselector file
- deselector where clause is commented out

There is the case where someone writes a deselector that removes everything (eg `WHERE id IS NOT NULL` or `WHERE 'hello' = 'hello'`)... you can't feasibly defend against that and it might be the easiest correct way purge everything from a site anyway.

pdowler (changes requested):
Final review after looking back at the story and task descriptions...

pdowler (approved):

PR #240, opened 2020-12-08
Cadc-8572 ringhold

jburke-cadc:
add info logging for processed artifact, fixed NullPointerException for missing config file


pdowler (changes requested):
What about the other bug on line 131 of Main?

pdowler (approved):

PR #241, opened 2020-12-09
Cadc-1166 ratik CADC-8837 CADC-8838

jburke-cadc:


ratik/src/main/java/org/opencadc/ratik/Main.java
```diff
@@ -0,0 +1,189 @@
...
+    private static final String CONFIG_FILE_NAME = "ratik.properties";
+    private static final String CONFIG_PREFIX = Main.class.getPackage().getName();
+    private static final String SQLGENERATOR_CONFIG_KEY = SQLGenerator.class.getName();
+    private static final String DB_SCHEMA_CONFIG_KEY = CONFIG_PREFIX + ".db.schema";
+    private static final String LOGGING_KEY = CONFIG_PREFIX + ".logging";
+    private static final String DB_URL_CONFIG_KEY = CONFIG_PREFIX + ".db.url";
+    private static final String DB_USERNAME_CONFIG_KEY = CONFIG_PREFIX + ".db.username";
+    private static final String DB_PASSWORD_CONFIG_KEY = CONFIG_PREFIX + ".db.password";
+    private static final String QUERY_SERVICE_CONFIG_KEY = CONFIG_PREFIX + ".queryService";
+    private static final String URI_BUCKETS_CONFIG_KEY = CONFIG_PREFIX + ".buckets";
```
pdowler, line 105: like fenwick, ratik needs to know if it is global or storage site to process several discrepancies correctly. In fenwick, that is done by configuring `trackSiteLocations`

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Validate local inventory.
+ */
+public class InventoryValidator implements Runnable {
+    private static final Logger log = Logger.getLogger(InventoryValidator.class);
+
+    private final ArtifactDAO artifactDAO;
+    private final URI resourceID;
+    private final String uriBuckets;
```
pdowler, line 95: The configured uriBuckets is a single bucket prefix or range of buckets so must be validated via BucketSelector. That probably belongs in main and pass the BucketSelector in the ctor.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
```diff
@@ -0,0 +1,146 @@
...
+    public InventoryValidator(Map<String, Object> daoConfig, URI resourceID, String uriBuckets) {
+        InventoryUtil.assertNotNull(InventoryValidator.class, "daoConfig", daoConfig);
+        InventoryUtil.assertNotNull(InventoryValidator.class, "resourceID", resourceID);
+        InventoryUtil.assertNotNull(InventoryValidator.class, "selector", uriBuckets);
+
+        this.artifactDAO = new ArtifactDAO(false);
+        this.artifactDAO.setConfig(daoConfig);
+        this.resourceID = resourceID;
+        this.uriBuckets = uriBuckets;
+
```
pdowler, line 113: Something missing here: ratik needs to mimic fenwick in selecting AllArtifacts or an ad-hoc filter policy. That was supported by a class named IncludeArtifacts. That means moving some code so it is re-usable here because it really does have to be the same and operators really have to put the same artifact-filter.sql file into fenwick and ratik or they will fight over what is to be included. This is needed for every discrepancy that mentions "filter policy".
* move code from fenwick to cadc-inventory-util (org.opencadc.inventory.util)
* update README for new config item and additional config file

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,560 @@
...
+    @Test
+    public void noDiscrepancy() throws Exception {
+        Calendar calendar = Calendar.getInstance();
+        Artifact artifact = new Artifact(URI.create("cadc:INTTEST/one.ext"), TestUtil.getRandomMD5(), calendar.getTime(), 1024L);
+        this.localEnvironment.artifactDAO.put(artifact);
+        this.remoteEnvironment.artifactDAO.put(artifact);
+
+        InventoryValidator testSubject = new InventoryValidator(this.localEnvironment.daoConfig, TestUtil.LUSKAN_URI, artifact.getBucket());
+        testSubject.run();
+
```
pdowler, line 125: for the test run, use of the uriBucket from the stored artifact would work but it should really be null to not complicate this test -- this also applicable to every other test.

**always use ArtifactDAO.get(UUID) (here and in other test methods) unless you really mean otherwise**. After a validate, the remote and local artifact should be identical, so the primary assert should be
```
Assert.assertEquals("metaChecksum", artifact.getMetaChecksum(), localArtifact.getMetaChecksum());
```
I should point out that ArtifactDAO.put(Artifact) does - as a side effect - assign metaChecksum and lastModified to the argument artifact. Putting the same one twice (lines 120-121) should obviously be fine and leave the artifact with the correct values (from the second put). It is intended to be done as a side effect so the values are available to the caller (eg for logging if desired) but it is subtle.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,560 @@
...
+     * after: Artifact not in L, DeletedStorageLocationEvent in L
+     */
+    @Test
+    public void explanation0_inL() throws Exception {
+        Calendar calendar = Calendar.getInstance();
+        Artifact artifact = new Artifact(URI.create("cadc:INTTEST/one.ext"), TestUtil.getRandomMD5(), calendar.getTime(), 1024L);
+        this.localEnvironment.artifactDAO.put(artifact);
+        this.remoteEnvironment.artifactDAO.put(artifact);
+
+        //.necessary to carefully generate a uri bucket, or will an out of range bucket work?
```
pdowler, line 150: filter policy excludes artifact means a config/artifact-filter.sql (like fenwick) containing a WHERE clause that does not match the remote artifact.

So, you do need an artifact-filter.sql file in {user.home}/config... should be in build/tmp like ringhold tests

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,560 @@
...
+        DeletedArtifactEvent deletedArtifactEvent = new DeletedArtifactEvent(artifact.getID());
+        this.remoteEnvironment.deletedArtifactEventDAO.put(deletedArtifactEvent);
+
+        InventoryValidator testSubject = new InventoryValidator(this.localEnvironment.daoConfig, TestUtil.LUSKAN_URI, artifact.getBucket());
+        testSubject.run();
+
+        Artifact localArtifact = this.localEnvironment.artifactDAO.get(artifact.getURI());
+        Assert.assertNull("local artifact found", localArtifact);
+
+        DeletedStorageLocationEvent dsle = this.localEnvironment.deletedStorageLocationEventDAO.get(artifact.getID());
```
pdowler, line 184: wrong event type

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,560 @@
...
+     * before: Artifact in L with R siteID in Artifact.siteLocations, not in R, DeletedStorageLocationEvent in R
+     * after: Artifact in not in L
+     */
+    @Test
+    public void explanation2_inL() throws Exception {
+        // case 1
+        UUID remoteSiteID = UUID.randomUUID();
+        UUID otherSiteID = UUID.randomUUID();
+
+        Calendar calendar = Calendar.getInstance();
```
pdowler, line 212: What's wrong with `new Date()`?


The SiteLocation is a reference to a StorageSite so the UUID is the StorageSite entity in the remoteEnvironment. That is normally put there by minoc (StorageSite is currently implicitly the resourceID of a "files" service) so it will not be there in these tests.

All the tests here are going to need to init the remoteEnvironment with a StorageSite when remote is storage because InventoryValidator is going to query luskan to get and use it (when L==global) just like fenwick does. See trackSiteLocations comment earlier.

Note: stopping review at this point because fixing this will likely cause other tests below to change but setup of the remote StorageSite may be needed above as well so I'll likely have to re-examine them.

pdowler (changes requested):

ratik/src/main/java/org/opencadc/ratik/Main.java
jburke-cadc: Added trackSiteLocations to ratik.properties.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
jburke-cadc: Moved BucketSelector creation into Main(), and pass that to the InventoryValidator constructor.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
jburke-cadc: Moved the ArtifactSelector classes into cadc-inventory-util and updated the README.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Using null for uriBucket in the InventoryValidator constructor, and using the Artifact id for a get.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Missed responding to these last comments. My uri bucket knowledge is still stuck on the AD site where the uri bucket IS the archive. When local is a storage site, the filter policy will include the storage site archives from global, and uri bucket could be empty to include all artifacts on the storage site. When local is global, the filter policy could be empty to include all storage site artifacts, and the uri bucket would somehow restrict the global artifacts to only those on the storage site. That last bit is what I'm not fully understanding yet. Need to look at global and a ceph site to see how that works.

Added support for the artifact-filter.sql config file.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Updated to the right event.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: I thought there might be a need to change the lastModifiedDate to return artifacts in a different order, so used a Calendar, but not needed so far. Changed to Date.

Added a StorageSite to the remote site before each test.

ratik/src/main/java/org/opencadc/ratik/Main.java
```diff
@@ -0,0 +1,202 @@
...
+    private static final String TRACK_SITE_LOCATIONS_CONFIG_KEY = CONFIG_PREFIX + ".trackSiteLocations";
+    private static final String ARTIFACT_SELECTOR_CONFIG_KEY = CONFIG_PREFIX + ".artifactSelector";
```
pdowler, line 109: For defined interface and plugin implementation, the key is the fully qualified class name of the interface and should be specified as code:
```
ArtifactSelector.class.getName()
```

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+     * after:  Artifact in L and R siteID in Artifact.siteLocations
+     */
+    @Test
+    public void explanation4_ArtifactNotInL_LocalIsGlobal() throws Exception {
+        UUID remoteSiteID = this.remoteEnvironment.storageSiteDAO.list().iterator().next().getID();
+
+        Artifact artifact = new Artifact(URI.create("cadc:INTTEST/one.ext"), TestUtil.getRandomMD5(),
+                                         new Date(), 1024L);
+        this.remoteEnvironment.artifactDAO.put(artifact);
+        this.localEnvironment.artifactDAO.put(artifact);
```
pdowler, line 642: For the artifact in the localEnvironment (global), I would add a different SiteLocation. reason: in global, there should always be at least 1 SiteLocation and I'd keep the test scenario consistent with that. It's possible to enforce that in the DAO and I might add that constraint at some point before we go to live production, so that could in theory break tests that don't do  setup right.

It also enables you to verify that the code did an "add siteID" rather than "set siteID"... implying that the local artifact goes from one to two so there should be another of the last assert to check that.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+            System.setProperty("user.home", USER_HOME);
+        }
+
+        Artifact localArtifact = this.localEnvironment.artifactDAO.get(artifact.getID());
+        Assert.assertNotNull("local artifact not found", localArtifact);
+        Assert.assertEquals("metaChecksum mismatch", artifact.getMetaChecksum(), localArtifact.getMetaChecksum());
+        Assert.assertTrue("artifact does not contains remote site location",
+                          localArtifact.siteLocations.contains(new SiteLocation(remoteSiteID)));
+    }
+
```
pdowler, line 660: For completeness: comments for explanation 6 and 7 here.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+    @Test
+    public void artifactUriCollision_LocalIsGlobal() throws Exception {
+        artifactUriCollision(true);
+    }
+
+    public void artifactUriCollision(boolean trackSiteLocations) throws Exception {
+        URI contentCheckSum = TestUtil.getRandomMD5();
+        UUID localID = UUID.randomUUID();
```
pdowler, line 684: In this discrepancy, the "pick winner" is not as simple as "remote wins", which is what you have here. You have to pick winner the same way that fenwick does because either fenwick or ratik could encounter this and they have to resolve it the same way.

So there are nominally two scenarios hiding here - one where the local artifact wins and one where the remote artifact wins - and the results are different. The checks need to verify that the expected winner ends up in L and the DAE is created for the loser. You don't have to explicitly check that the loser got deleted because DB constraints won't let the winner be inserted in that case

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+     * action: put Artifact
+     *
+     * explanation2: pending/missed artifact update in R
+     * evidence: ??
+     * action: do nothing
+     *
+     * before: Artifact in L & R, L metaChecksum != R metaChecksum
+     * after: Artifact in L with R metaChecksum
+     */
+
```
pdowler, line 725: I assume you were going to add two `@Test` methods that call this - it seems like a good idea to run these site-agnostic tests once for each type anyway (as above).

For this discrepancy, the `evidence: ??` for both explanations means we can't tell which is right. Unfortunately, here the actions are known and different so we still need to figure out what evidence can be used to resolve this. Right now I can't recall if I just forgot to go back and fill that in or if I thought about it and couldn't come up with a plan. Since I'll have to think about it again that's equivalent to having forgot to do it :smile: 

So, on the test setup:
* the fact that you create two artifacts (literally `new Artifact`) means that you have setup a collision and that resolution takes precedence over this one, which is about updates to existing artifacts -- in other tests where you created one Artifact and put it both environments it was actually important because they have the same Entity.id and that is what makes them "the same Artifact" (*not* having the same Artifact.uri)
* manually setting the metaChecksum in the localEnvironment makes that record in the database invalid, so that's a bad idea that you don't need

I think separate methods for the two explanations are needed here. For explanation1, I would create an Artifact and put it into localEnvironment and remoteEnvironment. Then modify the Artifact and put it to remote again. Now you have an update in R that has not been applied to L. Modify setup for explanation2.

This discrepancy could have been phrased as "pick a winner" with "no more evidence needed" and the  the action is "pick a winner, ...". For now, get the test setup right and I'll modify the document so that it is expressed that way or I'll modify the "collision" discrepancy so it has two explanations.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
```diff
@@ -0,0 +1,157 @@
...
+     * @param bucketSelector     uri buckets
+     * @param trackSiteLocations local site type
+     */
+    public InventoryValidator(Map<String, Object> daoConfig, URI resourceID, ArtifactSelector artifactSelector,
+                              BucketSelector bucketSelector, boolean trackSiteLocations) {
+        InventoryUtil.assertNotNull(InventoryValidator.class, "daoConfig", daoConfig);
+        InventoryUtil.assertNotNull(InventoryValidator.class, "resourceID", resourceID);
+        InventoryUtil.assertNotNull(InventoryValidator.class, "artifactSelector", artifactSelector);
+        InventoryUtil.assertNotNull(InventoryValidator.class, "bucketSelector", bucketSelector);
+        InventoryUtil.assertNotNull(InventoryValidator.class, "trackSiteLocations", trackSiteLocations);
```
pdowler, line 116: primitive `boolean` can't be null so the check on trackSiteLocations is not needed

pdowler (changes requested):

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+    public InventoryValidatorTest() throws Exception {
+
+    }
+
+    @Before
+    public void beforeTest() throws Exception {
+        writeConfig();
+        this.localEnvironment.cleanTestEnvironment();
+        this.remoteEnvironment.cleanTestEnvironment();
+        this.remoteEnvironment.initStorageSite(REMOTE_RESOURCEID);
```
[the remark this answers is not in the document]
jburke-cadc: Moved initStorageSite to only the tests that need a remote StorageSite.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+    public void explanation0_ArtifactInLocal_LocalIsStorage() throws Exception {
+        explanation0_ArtifactInLocal(false);
+    }
+
+    @Test
+    public void explanation0_ArtifactInLocal_LocalIsGlobal() throws Exception {
+        explanation0_ArtifactInLocal(true);
+    }
+
+    public void explanation0_ArtifactInLocal(boolean trackSiteLocations) throws Exception {
```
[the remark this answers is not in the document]
jburke-cadc: Had to read this about a dozen times before I understood the comment. Updated the Artifact URI to not match the filter policy.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+    }
+
+    /** discrepancy: artifact in L && artifact not in R
+     *
+     * explanation6: deleted from R, lost DeletedArtifactEvent
+     * evidence: ?
+     * action: assume explanation3
+     */
+    // if L == global then explanation3, L == storage then explanation1?
+    @Test
```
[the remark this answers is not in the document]
jburke-cadc: Updated the comments.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+    public void explanation6_ArtifactInLocal() throws Exception {
+        // L == global
+        // explanation3_ArtifactInLocal_LocalIsGlobal();
+        // L == storage
+        // explanation1_ArtifactInLocal_LocalIsStorage();
+    }
+
+
```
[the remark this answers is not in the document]
jburke-cadc: Added a comment.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+     * before: Artifact not in L, in R, DeletedStorageLocationEvent in L
+     * after: Artifact not in L, DeletedStorageLocationEvent in L
+     */
+    @Test
+    public void explanation2_ArtifactNotInLocal_LocalIsStorage() throws Exception {
+        Artifact artifact = new Artifact(URI.create("cadc:INTTEST/one.ext"), TestUtil.getRandomMD5(),
+                                         new Date(), 1024L);
+        this.remoteEnvironment.artifactDAO.put(artifact);
+
+        DeletedStorageLocationEvent remopteDeletedStorageLocationEvent = new DeletedStorageLocationEvent(artifact.getID());
```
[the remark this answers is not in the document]
jburke-cadc: Updated

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,759 @@
...
+                                                                    false);
+            testSubject.run();
+        } finally {
+            System.setProperty("user.home", USER_HOME);
+        }
+
+        Artifact localArtifact = this.localEnvironment.artifactDAO.get(artifact.getID());
+        Assert.assertNull("local artifact not found", localArtifact);
+
+        DeletedStorageLocationEvent localDeletedStorageLocationEvent =
```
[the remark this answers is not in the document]
jburke-cadc: Updated

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Added another SiteLocation to the local Artifact, and verify it's present after the validate has run.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Added the comments.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Spend a long time going through not many lines in fenwick that determine which Artifacts wins. Updated the tests to pick the same winner.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Yes forgot to add storage and global test. Separated this test into a test for each explanation.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
jburke-cadc: Removed the check.

ratik/README.md
```diff
@@ -13,21 +13,39 @@ org.opencadc.ratik.logging = {info|debug}
...
+
+# local site type, true = global site, false = storage site
+org.opencadc.ratik.trackSiteLocations={true|false}
 ```
 The range of uriBucket prefixes is specified with two values separated by a single - (dash) character; whitespace is ignored.
 
 ### cadcproxy.pem
 Remote calls will use this certificate to authenticate.
 
+### SQL includes
```
pdowler, line 38: put the file name here as the subsubsection and don't say `/config/artifact-filter.sql` in teh text. Look at the fenwick README for how the directory is specified up in the ##configuration subsection and follow that style

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,855 @@
...
+                                                                    new IncludeArtifacts(),null,
+                                                                    trackSiteLocations);
+            testSubject.run();
+        } finally {
+            System.setProperty("user.home", USER_HOME);
+        }
+
+        localArtifact = this.localEnvironment.artifactDAO.get(localArtifact.getID());
+        Assert.assertNull("local artifact found", localArtifact);
+
```
pdowler, line 729: The assert that localArtifact was deleted is part of the resolve (pick winner implied insert winning Artifact if it winner was in R). So this should also assert that remoteArtifact is now in localEnvironment.

I will update the doc.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,855 @@
...
+        URI contentCheckSum = TestUtil.getRandomMD5();
+        UUID localID = UUID.randomUUID();
+        UUID remoteID = UUID.randomUUID();
+        Calendar contentModifiedDate = Calendar.getInstance();
+
+        Artifact remoteArtifact = new Artifact(remoteID, URI.create("cadc:INTTEST/one.ext"), contentCheckSum,
+                                               contentModifiedDate.getTime(), 1024L);
+        this.remoteEnvironment.artifactDAO.put(remoteArtifact);
+
+        contentModifiedDate.roll(Calendar.YEAR, false);
```
pdowler, line 712: Since it is critical that the Artifact.uri values are equal, I would use remoteArtifact.getURI() in the localArtifact... or maybe a variable that both ctors use. Otherwise, everone has to read that string char-by-char and that's no fun :-)

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,855 @@
...
+
+    public void artifactChecksumMismatch_Explanation1(boolean trackSiteLocations) throws Exception {
+        UUID artifactID = UUID.randomUUID();
+        URI artifactURI = URI.create("cadc:INTTEST/one.ext");
+
+        Artifact artifact = new Artifact(artifactID, artifactURI, TestUtil.getRandomMD5(),
+                                              new Date(), 1024L);
+        this.localEnvironment.artifactDAO.put(artifact);
+        this.remoteEnvironment.artifactDAO.put(artifact);
+
```
pdowler, line 797: Since "pick a winner" is likely to be based on relative Entity.lastModified, I would put a short sleep in here so the updated artifact has a reliably different timestamp.... for someone testing locally with two local container database servers the clocks are the same so 50ms is plenty of sleep. (if the db servers are on different hosts, clocks can be off by ~minutes and that's a real pita to test)

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,855 @@
...
+                                                                    new IncludeArtifacts(),null,
+                                                                    trackSiteLocations);
+            testSubject.run();
+        } finally {
+            System.setProperty("user.home", USER_HOME);
+        }
+
+        Artifact localArtifact = this.localEnvironment.artifactDAO.get(artifact.getID());
+        Assert.assertNotNull("local artifact not found", localArtifact);
+        Assert.assertSame("local artifact metachecksum does not match",
```
pdowler, line 851: I prefer use of assertEquals... the check itself is sufficient, but subtle. This probably needs a comment like "local artifact is right and remote is wrong so discrepancy not be fixed"

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,855 @@
...
+        Log4jInit.setLevel("org.opencadc.inventory", Level.INFO);
+        Log4jInit.setLevel("org.opencadc.inventory.db", Level.INFO);
+        Log4jInit.setLevel("ca.nrc.cadc.db", Level.INFO);
+    }
+
+    static String INVENTORY_SERVER = "RATIK_TEST";
+    static String INVENTORY_DATABASE = "cadctest";
+    static String INVENTORY_SCHEMA = "inventory";
+    static String TMP_DIR = "build/tmp";
+    static String USER_HOME = System.getProperty("user.home");
```
pdowler, line 114: Maybe name this constant REMOTE_STORAGE_SITE - I was confused because in ratik "remote" normally means luskan so this only made sense when I looked at how it was used.

pdowler (changes requested):

ratik/README.md
jburke-cadc: Updated to follow the fenwick README style.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Fixed.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Cleaned up so you only have to read that string once!

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Hadn't thought of that, added a sleep between the artifact puts.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: Updated to use assertEquals and updated the assert message.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
jburke-cadc: That is a clearer name, updated the constant.

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -0,0 +1,852 @@
...
+    public void artifactUriCollision_LocalIsStorage() throws Exception {
+        artifactUriCollision(false);
+    }
+
+    @Test
+    public void artifactUriCollision_LocalIsGlobal() throws Exception {
+        artifactUriCollision(true);
+    }
+
+    public void artifactUriCollision(boolean trackSiteLocations) throws Exception {
```
pdowler, line 699: The setup for this test is still more complicated than it needs to be and hence to harder than necessary to verify correctness.

First, if you just do `new Artifact(URI, URI, Date)` it always assigns a new UUID. basically, `new Artifact` does exactly what it says: creates a new entity. Manually creating and assigning UUID is misleading because it looks like trying to finesse something different than that. (In CAOM, I never had that ctor at all and assigned entity.id via reflection... this was meant as a convenience for deserialisation code that other applications should never use)

Second, it would be more clear if you create the two test artifacts in the order implied by the scenario. So in case 1 create the localArtifact first then the remote.

Third, Calendar.roll(...) requires a javadoc lookup to see what true|false means. If you did something like this:
```
Date now = new Date();
Date older = new Date(now.getTime() - 2000L);
```
and then used those two values in the ctors it would be much more clear. I thought about now and "newer" but the simplest way to do that gives a contentLastModified in the future and that seems like a strange thing that future validation constraints might reject. I like test scenarios to be normal/realistic.

Given the complexity/subtlety of the logic in this test class it's important to not have distracting

pdowler (changes requested):

pdowler (approved):

PR #247, opened 2021-01-14
CADC-1166-ratik CADC-8840 iterator comparison logic

jburke-cadc:


pdowler:
Two critical things are missing from the code so far:
1. use of the selected bucket(s) in the queries; this will be part of the ArtifactDAO API for local and needs to be used in the remote (TAP) query as well; currently the usage is not great because you get an iterator and need an extra loop (eg "all" means iterating from 0 to f -- a loop of 16 steps) when it is feasible to write a single condition in the query to get a range of buckets). So, I'll see if I can figure out a good way to handle that when doing the db-lib task so we can formulate one query rather than add another layer of while loop... 
2. queries must "order by" so the iterators are in the same order; it remains TBD if the order is by Artifact.id or Artifact.uri but it probably has to be the latter because Artifact.uriBucket is computed from Artifact.uri (so a uri always ends up in the same bucket)

Once you consider #2, the test content will have to be modified to take ordering into account; that means picking suitably ordered Artifact.uri values in each case. I would also add an INFO level log message in the test validate methods to print out the two args -- nothing better than a visual like that to show that the method is being called as you expect (number of times and pair of args).

jburke-cadc:
1. Wouldn't the artifact-filter.sql used in the TAP query be expected to return the same range of Artifacts as defined by the bucket range? Or is that putting too much onus on the user of ratik to give a bucket range and artifact-filter.sql query that return the same results, and it's better for ratik to generate the TAP query from the bucket range.

jburke-cadc:
2. Okay I'll use Artifact.uri to sort the query results, update the tests, and add the INFO log message with the validate method arguments.

pdowler:
1. No, the artifact-filter is something like `WHERE uri like 'cadc:CFHT/% ...` and specifies the set of artifacts are supposed to be in the local site. The buckets are a "random" subset of that set and are used to validate arbitrary subsets of the whole (for robustness). The arbitrary buckets could be validated in parallel (different instances of ratik) and if validation fails you only have to redo that bucket.

jburke-cadc:
So the bucket selector is a further constraint on the artifact-filter, and both are used to generate the remote query?

If the artifact-filter.sql is `WHERE uri like 'cadc:CHFT/%'` 
and the bucket range is `9` to `b`,
 the WHERE would be:
`WHERE uri like 'cadc:CFHT/%' AND uribucket = '9' AND uribucket = 'a' and uribucket = 'b'`?

pdowler:
Yes, but you'll have to be careful to put the constraint from artifact-filter into brackets, eg:
```
WHERE (uri like 'cadc:CFHT/%') AND (uribucket LIKE 'a%')
```
where `a` is a value returned by BucketSelector. If the bucket config was something like a-c (all buckets starting with a, b , or c) one could either do 3 queries as above. 

If you knew the alphabet of uriBucket values (lower case hex) or figure out how to write the ADQL as 
```'a' <= uriBucket AND uriBucket < 'd'``` 
(since c is included but d is not). If you happened to know the alphabet and length of uriBucket values you could write 
```'a0000' <= uriBucket AND uriBucket < 'cffff'```
(alphabet is is hard coded to lower case hex and length 5). I have not thought through whether/how that specific info could be expressed in the API and not simply hard coded or assumed. Think about that vs looping with BucketSelector.getBucketIterator and we can discuss pros and cons face to face.

**Comment by jburke-cadc** [comment]
Updated the remote TAP query to ORDER BY uri ASC. The comparison to determine if a local Artifact is considered before or after a remote Artifact returns the same ordering. Although how lower and upper case characters are ordered was not what I was expecting. Thankfully PostgreSQL and Java use the same ordering.
The InventoryValidator now uses an ArtifactValidator class to do the local and remote Artifact comparison. Also helps separate the iterator and logic pull requests.

**Comment by jburke-cadc** [comment]
I had to rework the InventoryValidatorIteratorsTest int test, which checks the ordering when comparing the local and remote sets of Artifacts. Given you can't set the bucket for an Artifact, you can't get a known set of Artifacts for a bucket. I overwrote the methods that return the local and remote iterators to return known sets of Artifacts. After a couple of tweaks to the InventoryValidator that int test passes.

File: ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java [hunk 1]
```diff
@@ -150,7 +169,215 @@ public InventoryValidator(Map<String, Object> daoConfig, URI resourceID, Artifac
...
+        log.debug(String.format("order artifact uri's:\n local - %s\nremote - %s",
+                                localArtifact == null ? "null" : localArtifact.getURI(),
+                                remoteArtifact == null ? "null" : remoteArtifact.getURI()));
+        int result;
+        if (localArtifact == null) {
+            result = 1;
+        } else if (remoteArtifact == null) {
+            result = -1;
+        } else {
+            result = localArtifact.getURI().toString().compareTo(remoteArtifact.getURI().toString());
```
pdowler, line 270: URI implements Comparable so you don't need to convert to string

This method that is easily testable with just code: please add unit test for the 5 scenarios: 
```
a,null 
null,b 
a,a
a,b 
b,a
```

Since this method is able to order non-null vs null, you need to state whether null is before or after non-null. The common thing is for null to come after non-null but I think you are doing the opposite here... it doesn't really matter as long as the conditionals up at line ~224+ do the same thing, but since we do have examples of "NULLS LAST" ordering elsewhere it's something to think about. It must be explicit.

re: comment about "ORDER BY uri ASC": this comment from cadc-inventory-db library when ordering by StorageLocation is related: 
```
// NOTE: StorageLocation.compare() specifies the correct order
// null storageBucket to come after non-null
// postgresql: the default order is equivalent to explicitly specifying ASC NULLS LAST
// default behaviour may not be db-agnostic
```

aside: if you want the string form of a URI (eg to serialise), always use toASCIIString(). toString() is a default method for debugging

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
```diff
@@ -150,7 +169,215 @@ public InventoryValidator(Map<String, Object> daoConfig, URI resourceID, Artifac
...
+     * @return  String query.  Never null.
+     */
+    String buildRemoteQuery(final String bucket)
+        throws ResourceNotFoundException, IOException {
+        final StringBuilder query = new StringBuilder();
+        query.append("SELECT id, uri, contentChecksum, contentLastModified, contentLength, contentType, ")
+            .append("contentEncoding, lastModified, metaChecksum FROM inventory.Artifact ");
+
+        if (StringUtil.hasText(this.artifactSelector.getConstraint())) {
+            log.debug(String.format("\nInjecting artifact selector clause '%s'\n",
```
pdowler, line 339: rather than these two debug messages just log the entire query somewhere at debug level; that is always more useful. I would be more inclined to put that in getRemoteIterator... oh wait: it is already there... so these debug statements here are just noise.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
```diff
@@ -150,7 +169,215 @@ public InventoryValidator(Map<String, Object> daoConfig, URI resourceID, Artifac
...
+                                remoteArtifact == null ? "null" : remoteArtifact.getURI()));
+        int result;
+        if (localArtifact == null) {
+            result = 1;
+        } else if (remoteArtifact == null) {
+            result = -1;
+        } else {
+            result = localArtifact.getURI().toString().compareTo(remoteArtifact.getURI().toString());
+        }
+        if (log.isDebugEnabled()) {
```
pdowler, line 272: This debugging isn't useful - the unit test of this method would be. debug messages in there would be fine if they help

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
```diff
@@ -150,7 +169,215 @@ public InventoryValidator(Map<String, Object> daoConfig, URI resourceID, Artifac
...
+                    if (localArtifact == null && remoteArtifact == null) {
+                        artifactsToValidate = false;
+                        continue;
+                    }
+                    log.debug(String.format("comparing Artifacts:\n local - %s\nremote - %s",
+                                            localArtifact, remoteArtifact));
+
+                    // check if Artifacts are the same, or if the local Artifact
+                    // precedes or follows the remote Artifact.
+                    int order = orderArtifacts(localArtifact, remoteArtifact);
```
pdowler, line 223: I would use a switch statement here: explicitly -1, 0, 1, and then `default` being a failure

pdowler (changes requested):

pdowler:
With respect to "I had to rework the InventoryValidatorIteratorsTest int test ... you can't get a known set of Artifacts for a bucket."

Yeah, the ArtifactDAO methods all allow a null bucketPrefix and getting all artifacts. The buckets part of the models and APIs are there to support picking an arbitrary fraction of the artifacts to work on so that one can sibdivide the work in order to do smaller chunks and make progress or to do work in parallel (multiple instances of ratik, tantar, etc). That is completely orthogonal to the correctness of validating a set of artifacts, so it would be nice to use null bucketPrefix for most of the tests and then just have 1-2 tests that verify that bucketPrefix is used correctly in both queries. However, the ctor for InventoryValidator and the BucketSelector don't allow for that. 

BucketSelector selector could have an ALL concept? 
InventoryValidator should have a package-access ctor for test use that allows a null BucketSelector?

jburke-cadc:
I saw that the DAO allowed a null bucketPrefix, but the BucketSelector did not. I don't know if an 'all buckets' concept would be useful outside of testing, but I think it would a worth while addition.
A no arg BucketSelector constructor could mean 'all buckets', or a BucketSelector.ALL constant to pass to the constructor. The constant arg makes it clear what you are doing. ALL sounds like a bucket range from 0 - f though, but the docs would explain how it is meant to be used.

Would other apps using BucketSelector need to updated to expect an ALL argument? A quick search shows tartar and critwall use the BucketSelector. 

You could do a lot of hacky things with an InventoryValidator to set a null bucketSelector. Then update the bucket looping to handle a null bucketSelector. That at least keeps the changes in a single class.

Adding the ALL concept to BucketSelector seems the proper way to go, but does it make sense for other apps. tantar can use a single bucket, the archive. I guess you could look for the ALL bucket and throw an IllegalArgumentException or UnsupportedOperationException. Possibly the same with critwall.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -0,0 +1,106 @@
...
+ ************************************************************************
+ */
+
+package org.opencadc.ratik;
+
+import java.net.URI;
+import org.apache.log4j.Logger;
+import org.opencadc.inventory.Artifact;
+import org.opencadc.inventory.db.ArtifactDAO;
+
```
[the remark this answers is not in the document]
jburke-cadc: Updated the class java doc.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
jburke-cadc: The null vs non-null behaviour was backwards to how it is handled else where. Updated to follow the same pattern where null follows non-null.
Added a unit test for the Artifact ordering method.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
jburke-cadc: Removed the two debug messages.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
jburke-cadc: Removed the debug message. Doesn't help for unit tests.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
jburke-cadc: The orderArtifacts(local, remote) method now returns -1,0,1 instead of <neg int>,0,<pos int>, for use with a switch statement.

pdowler (approved):

pdowler (approved):

PR #250, opened 2021-01-26
Cutout stream

at88mph:
- Integration tests for Minoc
   - Tests will download from public Vault into the caller's `.config` folder for local compare.  After download, subsequent tests will use those.
 - Add `FitsTest` file for Fits comparisons
 - Modified GetAction to use the DAL FitsOperations class to cutout to the stream.
   - SODA enabled Minoc


minoc/src/main/java/org/opencadc/minoc/GetAction.java
```diff
@@ -99,12 +113,30 @@ public void doAction() throws Exception {
         
         Artifact artifact = getArtifact(artifactURI);
         HeadAction.setHeaders(artifact, syncOutput);
-        
+
+        final List<String> requestedSubs = syncInput.getParameters(SodaParamValidator.SUB);
```
pdowler, line 117: I didn't think you would need to do this but SyncInput doesn't expose the whole param map. I had planned on something like 
```
List<ExtensionSlice> slices = SODA_PARAM_VALIDATOR.validateSUB(syncInput.getParameterMap());
```
because this is going to get ugly when you also need to do SODA and you'll want to use 
`SODA_PARAM_VALIDATOR.validateAllShapes(Map)` without having to know all the params that carry shape values.... the idea with the validators is that you let them look through the whole map and pick out the things they know about. ugh.

Ok  for now, but I'll create an issue to add SyncInput.getParameterMap() and use it here

minoc/src/main/java/org/opencadc/minoc/GetAction.java
```diff
@@ -99,12 +113,30 @@ public void doAction() throws Exception {
         
         Artifact artifact = getArtifact(artifactURI);
         HeadAction.setHeaders(artifact, syncOutput);
-        
+
```
pdowler, line 116: If a cutout is being performed, most of the header content changes. The above `HeadAction.setHeaders(...)` call is going to set wrong values in that case. So, that line has to move into the plain get output block (~line 124) and the cutout block should only set Content-Disposition and Content-Type (we'll worry about ops that change type when it happens).

There should be an intTest that verifies headers and it should assert that Content-Length, Content-MD5, and (probably) Content-Encoding are all not set for a cutout. And Digest, even though we don't currently set it (there's an issue for that).

pdowler (changes requested):

at88mph:
Thanks.  Rework submitted.

pdowler (approved):

PR #253, opened 2021-01-29
Fixed the Storage URI

andamian:
CADC-9223


cadc-storage-adapter-ad/src/test/java/org/opencadc/inventory/storage/ad/AdStorageQueryTest.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.Date;
+import org.apache.log4j.Level;
+import org.apache.log4j.Logger;
+import org.junit.Assert;
+import org.junit.Test;
+
+import org.opencadc.inventory.storage.StorageMetadata;
+import org.opencadc.tap.TapRowMapper;
+
+
```
pdowler, line 83: The catch with query and RowMapper is always whether the query select list and row mapper agree on column order. This would be much better to include in the intTest and process a real query result where uri != inventoryURI.

relevant intTest is **AdStorageAdapterIteratorTest.testIteratorOrder** that queries IRIS... if you cvan add some asserts to that to verify that the mapper is getting the two URIs in the right place, that would be better than this whole test.

Probably just  in the first loop (line ~164) just check the scheme(s) are `ad` in the storageID and `cadc` in the artifactURI.

pdowler (changes requested):

pdowler (approved):

PR #255, opened 2021-02-03
Cadc-1166 CADC-8841 Artifact validation logic

jburke-cadc:
Draft pull request for the ArtifactValidator class. It looks for evidence of a discrepancy between the two Artifacts, and logs the evidence found and action required.


ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+        // explanation3: L==storage, new Artifact in R, pending/missed new Artifact event in L
+        // also
+        // explanation0: filter policy at L changed to include artifact in R
+        // explanation6: deleted from L, lost DeletedArtifactEvent
+        // explanation7: L==storage, deleted from L, lost DeletedStorageLocationEvent
+        // evidence: ?
+        // action: insert Artifact
+        log.debug("checking explanation 3");
+        if (!this.trackSiteLocations) {
+            Artifact local = this.artifactDAO.get(remote.getID());
```
pdowler, line 269: you already know the artifact is not in L by being in the validateRemote method, so nothing to check

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+                log.info(String.format("Artifact not in L && in R - Explanation0,3,6,7\naction: insert Artifact\n"
+                                            + "remote - %s",
+                                        remote));
+                //TODO do action
+                return;
+            }
+        }
+
+        // explanation4: L==global, new Artifact in R, pending/missed changed Artifact event in L
+        // evidence: Artifact in local db but siteLocations does not include remote siteID
```
pdowler, line 280: Now that I see some code, this one doesn't quite make sense... I will modify the docs to clarify.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+        // evidence: no more evidence needed
+        // action: pick winner, create DeletedArtifactEvent for loser, delete loser if it is in L, insert winner if winner was in R
+        log.debug("checking artifact.id mismatch");
+        if (!local.getID().equals(remote.getID())) {
+            log.info(String.format("ID mismatch:\npick winner\nuri - %s\n local - %s %s\nremote - %s %s",
+                                   local.getURI(),
+                                   local.getID(), dateFormat.format(local.getLastModified()),
+                                   remote.getID(), dateFormat.format(remote.getLastModified())));
+            if (local.getContentLastModified().before(remote.getContentLastModified())) {
+                //TODO do action
```
pdowler, line 319: The log message outside the missing logic is goiong to be just noise: one log.info per action taken!

There is always some action here; please complete the logic and log the actual actions that would be taken eg:
```
DeletedArtifactEvent dae = new DeletedArtifactEvent(loser.getID());
log.info("create: " + dae + " " + loser.getURI() + " reason: ???");
log.info("delete: " + {loser ident} + " reason: ???");
...
```

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+        }
+
+        // discrepancy: artifact in both && valid metaChecksum mismatch
+        log.debug("checking valid metaChecksum mismatch");
+        if (!local.getMetaChecksum().equals(remote.getMetaChecksum())) {
+            log.info(String.format("MetaChecksum mismatch:\n uri - %s %s\n local - %s %s\nremote - %s %s",
+                                   local.getURI(), local.getID(), local.getMetaChecksum(),
+                                   dateFormat.format(local.getLastModified()), remote.getMetaChecksum(),
+                                   dateFormat.format(remote.getLastModified())));
+            // explanation1: pending/missed artifact update in L
```
pdowler, line 331: as above: complete the logic and log the actions taken

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+            // explanation2: pending/missed artifact update in R
+            // evidence: ??
+            // action: do nothing
+            //TODO determine action
+            return;
+        }
+
+        // Local & remote Artifact's have the same ID.
+        if (local.getID().equals(remote.getID())) {
+            log.info(String.format("Valid: - %s", local));
```
pdowler, line 345: here and other places: if no action taken do not log (even logging at debug will make debug too noisy to be useful)... 

You __could__ have something like `private void logNoAction(String msg, ...)` with some var-args where the actual method body is normally comented out...

pdowler (changes requested):

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+        }
+    }
+
+
+    /**
+     * Get a remote Artifact
+     */
```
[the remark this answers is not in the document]
jburke-cadc: Copied the ArtifactMapper to ArtifactValidator and forgot to remove the original.
I moved the RowMapper classes into org.opencadc.inventory.query in cadc-inventory-util. You need a matching query for the RowMapper so added a method to get the query.
ratik & fenwick use the query and mapper for a TAP query, so added a method that takes a where clause and resourceID, and returns the query results. Could change resourceID to TapClient, since the resourceID doesn't change for a ratik or fenwick instance.
ratik doesn't need the StoreageSite query if it can do a local query, but added it if useful in general? If not will delete. And for completeness could add a class for ObsoleteStorageLocation if that would be useful.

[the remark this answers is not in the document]
pdowler: This is not what I suggested would be useful and re-usable code. You really only need the ArtifactRowMapper and a string constant with the base query. Everything else is not useful because different applications will do something different. For example, ratik needs to compare artifact streams so it does `order by uri ` while fenwick only does incremental harvesting so it ... does something else.  

cadc-inventory-utll should have ArtifactRowMapper as a normal class. It could have the base query, eg:
```
public static final String BASE_QUERY = "SELECT ... FROM inventory.Artifact";
```
but that does hard-code the schema. I expect that's OK because even though the schema is configurable anywhere (except luskan) I expect that even if it was the solution would be to always use `inventory` in TAP and have luskan re-write the query... because of that it's probably going to be OK to hard code the schema, but it bears considering the consequences.

Same applies to DeletedArtifactEventMapper and DeletedStorageLocationMapper; code is not at the point where I can tell if both fenwick and ratik need a StorageSiteMapper  so leave that in ratik.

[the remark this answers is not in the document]
jburke-cadc: As above, created RowMapper classes for Artifact, DeletedArtifactEvent, and DeletedStorageLocationEvent, with the base query for each, and updated ratik to use the new classes.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+        return returned;
+    }
+
+    /**
+     * Get a remote DeletedArtifactEvent
+     */
+    DeletedArtifactEvent getRemoteDeletedArtifactEvent(UUID id)
+        throws InterruptedException, IOException, ResourceNotFoundException, TransientException {
+        final TapClient<DeletedArtifactEvent> tapClient = new TapClient<>(this.resourceID);
+        final String query = String.format("SELECT id, lastModified, metaChecksum FROM inventory.DeletedArtifactEvent "
```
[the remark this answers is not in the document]
jburke-cadc: Updated the queries with tick marks where needed.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+    }
+
+    /**
+     * Get a remote DeletedStorageLocalEvent
+     */
+    DeletedStorageLocationEvent getRemoteDeletedStorageLocationEvent(UUID id)
+        throws InterruptedException, IOException, ResourceNotFoundException, TransientException {
+        final TapClient<DeletedStorageLocationEvent> tapClient = new TapClient<>(this.resourceID);
+        final String query = String.format("SELECT id, lastModified, metaChecksum "
+                                               + "FROM inventory.DeletedStorageLocationEvent WHERE id=%s", id);
```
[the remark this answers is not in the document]
jburke-cadc: Updated

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+            tapClient.execute(query, new DeletedStorageLocationEventRowMapper());
+        if (results.hasNext()) {
+            returned = results.next();
+        }
+        return returned;
+    }
+
+    /**
+     * Get the StorageSite for the remote instance resourceID;
+     */
```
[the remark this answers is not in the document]
jburke-cadc: Removed the StorageLocation RowMapper and methods.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -96,15 +120,383 @@ public ArtifactValidator(ArtifactDAO artifactDAO, URI resourceID, boolean trackS
...
+    protected void validateLocal(Artifact local)
+        throws InterruptedException, IOException, ResourceNotFoundException, TransientException {
+
+        // explanation0: filter policy at L changed to exclude artifact in R
+        // evidence: Artifact in R without filter
+        // action: delete Artifact, if (L==storage) create DeletedStorageLocationEvent
+        log.debug("checking explanation 0");
+        Artifact remote = getRemoteArtifact(local.getURI());
+        if (remote != null) {
+            log.info(String.format("Artifact in L && not in R - Explanation0: Artifact in R without filter\n"
```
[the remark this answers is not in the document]
jburke-cadc: Simplified the info level logging for the explanations.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
jburke-cadc: Do not check for the local artifact for that explanation.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
jburke-cadc: Updated log messages.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
jburke-cadc: Updated

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
jburke-cadc: Added a method for logging when no action is necessary where the logging can be uncommented for dev work.

ratik/src/main/java/org/opencadc/ratik/Main.java
```diff
@@ -169,17 +170,18 @@ public static void main(final String[] args) {
             final String configuredQueryService = props.getFirstPropertyValue(QUERY_SERVICE_CONFIG_KEY);
             final URI resourceID = URI.create(configuredQueryService);
 
+            final String configuredFileService = props.getFirstPropertyValue(FILE_SERVICE_CONFIG_KEY);
+            final StorageSite remoteSite = new StorageSite(URI.create(configuredFileService), configuredFileService,
```
pdowler, line 174: The change in InventoryValidator to have a ref to the remote StorageSite (instead of a boolean) did not mean that this should propagate all the way to main and config file content. In fact, the config should be as it was (and like fenwick) because making the remote files service configurable just increases the chance of misconfiguring ratik. Plus, how would one configure for running ratik in a storage site and validating against a remote global?

The config structure was correct already. Like fenwick, ratik has to query for and assign the StorageSite **if trackSiteLocations  is true** and leave it null if false.

pdowler (changes requested):

ratik/src/main/java/org/opencadc/ratik/Main.java
jburke-cadc: Removed the files resourceID from the config file, and query the remote site for the StorageSite if running on global.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
pdowler: On seciond thought, this explanation is OK when you also consider that set L is filtered by (remote) siteID

pdowler (commented):
ArtifactValidator 

line~263: why do you need to get the local artifact by ID? In all the explanations you know the local artifact is not there. For the log message, instead of "reasons: multiple" just put "reason: pending/missed Artifact". If we could prove "lost deletion" then we'd take the opposite action, so it isn't the case that there are multiple reasons to keep the artifact.

line~303: log message should say something like "resolve Artifact.id collision: ...". There are 1-2 actions here and they require separate log messages. Always: create DeletedArtifactEvent for loser. Sometimes: delete loser.

line~313: you still need to write the logic to differentiate the two cases and then put the log message inside where it specified the action and reason. I have updated the `evidence` in my own repo (https://github.com/pdowler/storage-inventory/blob/master/storage-inventory-dm/README.md) to help with that.

Side note: for the metaChecksum discrepancy, it explicitly says `valid` metaChecksum: that means that when ratik gets an Artifact it must verify the checksum before doing anything else: compute a new value of the checksum and compare it to the one in the Artifact. If those mismatch, it means some part of the state was not serialised faithfully. This validation is needed in both fenwick and ratik and probably means a complete fail and abort. In fenwick, InventoryValidator line~500.

pdowler (approved):

PR #258, opened 2021-02-05
Add "files" end point to the maven service

andamian:
CADC-9269

### DIFF
--- a/raven/src/main/java/org/opencadc/raven/FilesAction.java
+++ b/raven/src/main/java/org/opencadc/raven/FilesAction.java
@@ -3,7 +3,7 @@
 *******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
 **************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
 *
-*  (c) 2020.                            (c) 2020.
+*  (c) 2021.                            (c) 2021.
 *  Government of Canada                 Gouvernement du Canada
 *  National Research Council            Conseil national de recherches
 *  Ottawa, Canada, K1A 0R6              Ottawa, Canada, K1A 0R6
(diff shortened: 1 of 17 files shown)


raven/src/main/java/org/opencadc/raven/FilesAction.java
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Create a valid artifact uri.
+     * @param uri The input string.
+     * @return The artifact uri object.
+     */
+    private URI createArtifactURI(String uri) {
+        try {
+            log.debug("artifactURI: " + uri);
+            artifactURI = new URI(uri);
```
pdowler, line 125: This is assigning to the member var and returning it for another assignment (after checking). Use a local variable here instead of assignment side effect.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -0,0 +1,234 @@
...
+/**
+ *
+ * @author adriand
+ */
+public class RavenInitAction extends InitAction {
+    private static final Logger log = Logger.getLogger(RavenInitAction.class);
+
+    // config keys
+    private static final String RAVEN_KEY = "org.opencadc.raven";
+    static final String RESOURCE_ID_KEY = RAVEN_KEY + ".resourceID";
```
pdowler, line 92: raven doesn't need to know it's own resourceID

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -0,0 +1,234 @@
...
+        
+        try {
+            this.resourceID = new URI(rid);
+            log.info("initConfig: OK");
+        } catch (URISyntaxException ex) {
+            throw new IllegalStateException("invalid config: " + RESOURCE_ID_KEY + " must be a valid URI");
+        }
+    }
+
+    private void initStorageSite() {
```
pdowler, line 183: raven is not a StorageSite definitely does not put  a StorageSite record into the database. The database pool props try to clearly state that this is a query pool (like the query pool in luskan) and thus is a read-only pool. You must have configured with the wrong account to make this work, but it doesn't belong here at all and the pool should use a read-only account (tapuser in the cadc-postgresql-dev container)

raven/src/main/java/org/opencadc/raven/ServiceAvailability.java
```diff
@@ -125,22 +127,27 @@ public AvailabilityStatus getStatus() {
...
             // check for a certficate needed to perform network ops
             CheckCertificate checkCert = new CheckCertificate(SERVOPS_PEM_FILE);
             checkCert.check();
             log.info("cert check ok");
 
-            CheckResource checkResource;
-            URI credURI = localAuthority.getServiceURI(Standards.CRED_PROXY_10.toString());
-            url = reg.getServiceURL(credURI, Standards.VOSI_AVAILABILITY, AuthMethod.ANON).toExternalForm();
-            checkResource = new CheckWebService(url);
+            String url = "https://localhost.cadc.dao.nrc.ca/baldur/availability";
```
pdowler, line 135: What is this? It can' stay in here like this... there are potentially multiple grant providers to call and (for reasons) I would not make this check depend on any of them.

revert changes to this file

pdowler (changes requested):

raven/src/main/java/org/opencadc/raven/FilesAction.java
andamian: Doh! Fixed

raven/src/main/java/org/opencadc/raven/ServiceAvailability.java
andamian: This is garbage from early debugging of my local system. I was pretty sure I've reverted, not sure how it got in. Fixed now.

pdowler (approved):

PR #268, opened 2021-02-23
CADC-8679 update minoc to use Digest HTTP header instead of Content-MD5

jburke-cadc:
open-cadc/core branch CADC-8679 contains the cadc-util changes to support the Digest header

### DIFF
--- a/minoc/src/intTest/java/org/opencadc/minoc/IncorrectPutMetadataTest.java
+++ b/minoc/src/intTest/java/org/opencadc/minoc/IncorrectPutMetadataTest.java
@@ -67,6 +67,7 @@
 
 package org.opencadc.minoc;
 
+import ca.nrc.cadc.net.Digest;
 import ca.nrc.cadc.net.HttpDelete;
 import ca.nrc.cadc.net.HttpGet;
 import ca.nrc.cadc.net.HttpTransfer;
@@ -135,7 +136,7 @@ public class IncorrectPutMetadataTest extends MinocTest {
                     HttpUpload put = new HttpUpload(in, artifactURL);
                     URI checksumURI = computeChecksumURI(incorrectData.getBytes());
                     String algorithm = checksumURI.getScheme();
-                    String checksum = HttpTransfer.base64Encode(checksumURI.getSchemeSpecificPart());
+                    String checksum = Digest.base64Encode(checksumURI.getSchemeSpecificPart());
                     put.setRequestProperty(HttpTransfer.DIGEST, String.format("%s=%s", algorithm, checksum));
                     put.setRequestProperty(HttpTransfer.CONTENT_LENGTH, Long.toString(bytes.length));
                     put.run();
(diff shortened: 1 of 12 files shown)


minoc/src/main/java/org/opencadc/minoc/PutAction.java
```diff
@@ -136,37 +136,30 @@ public Content accept(String name, String contentType, InputStream inputStream)
     public void doAction() throws Exception {
         
         initAndAuthorize(WriteGrant.class);
-        
-        String md5Header = syncInput.getHeader("Content-MD5");
+
+        String digestHeader = syncInput.getHeader("Digest");
```
pdowler, line 140: SyncOutput.setDigest(URI) exists... why not add `URI SyncInput.getDigest()` to match what clients have in `HttpTransfer`? Should not be using DigestUtil here

minoc/src/intTest/java/org/opencadc/minoc/IncorrectPutMetadataTest.java
```diff
@@ -304,7 +308,10 @@ public Object run() throws Exception {
                     // put file
                     InputStream in = new ByteArrayInputStream(bytes);
                     HttpUpload put = new HttpUpload(in, artifactURL);
-                    put.setRequestProperty(HttpTransfer.CONTENT_MD5, computeMD5(bytes));
+                    URI checksumURI = computeChecksumURI(bytes);
+                    String algorithm = checksumURI.getScheme();
+                    String checksum = DigestUtil.base64Encode(checksumURI.getSchemeSpecificPart());
+                    put.setRequestProperty(HttpTransfer.DIGEST, String.format("%s=%s", algorithm, checksum));
```
pdowler, line 314: Ack! `setDigest(URI)` ???

pdowler (changes requested):

minoc/src/main/java/org/opencadc/minoc/PutAction.java
jburke-cadc: Agree, added URI getDigest() to SyncInput.

minoc/src/intTest/java/org/opencadc/minoc/IncorrectPutMetadataTest.java
jburke-cadc: HttpUpload has deprecated the individual set[Header] methods and uses setRequestProperty(name, value). But most header values are known and don't require computation, unlike the Digest header. Worthwhile to add a setDigest(URI checksum) to HttpUpload?

pdowler: This should use the `setDigest(URI)` method instead of re-implementing the http-specific encoding junk here. That method should be available for any method that sends body content (HttpPost and HttpPut)... whenever there is some code to receive something and provide it with a get method, there is probably a place for a symmetric set method to send such a value.

pdowler (approved):

PR #270, opened 2021-02-24
CADC-8667 update tantar to cleanup ObsoleteStorageLocations

jburke-cadc:


tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
```diff
@@ -666,4 +673,44 @@ public Artifact next() {
...
+                LOGGER.debug("start transaction...");
+                transactionManager.startTransaction();
+                LOGGER.debug("start transaction... OK");
+
+                this.obsoleteStorageLocationDAO.delete(obsoleteStorageLocation.getID());
+                LOGGER.debug(String.format("deleted %s", obsoleteStorageLocation));
+
+                LOGGER.debug("commit transaction...");
+                transactionManager.commitTransaction();
+                LOGGER.debug("commit transaction... OK");
```
at88mph, line 699: This looks like it will still get executed in `reportOnly` mode.  Does this method need a `canTakeAction()` check?

at88mph (changes requested):

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
jburke-cadc: Yes it does need to check canTestAction() before executing. Updated.

at88mph (dismissed):

pdowler (changes requested):
The issue had a fairly detailed description of what "consult ObsoleteStorageLocation" meant and exactly how that should work. Specifically, it should happen before a StorageMetadata object is returned and available for comparison. The implementation here checks for an OSL, maybe deletes stuff, but then proceeds to compare the (now possibly non-existent) StorageMetadata object to an Artifact anyway.

pdowler:
This should not visibly happen in the double iterator merge-compare loop. It should happen inside the storage iterator and such a StorageMetadata should not be returned from that iterator at all.

The problem in the PR code is that after `checkObsoleteStorageLocation(storageMetadata)` the code proceeds to compare that storageMetadata object to an artifact. That's wrong. The storageMetadata instance may no longer represent a file in storage. You can't set it to null without getting an NPE on line 302... So what happens next? Depends on what kind of StorageAdapter and ResolutionPolicy is in use... which is why the approach is wrong.

jburke-cadc:
Moved the ObsoleteStorageLocation check into the iterator hasNext() method. If the StorageLocation is obsolete and the file deleted, can try the next StorageMetadata item in the iterator, until a StorageMetadata with a not-obsolete storage location is found and returned.

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
```diff
@@ -666,4 +676,54 @@ public Artifact next() {
             }
         };
     }
+
+    /**
+     * Check the StorageMetadata for a matching ObsoleteStorageLocation and if found
+     * delete the file from storage and delete the ObsoleteStorageLocation.
+     */
+    boolean isObsoleteStorageLocation(StorageMetadata storageMetadata) {
+        if (canTakeAction()) {
```
pdowler, line 685: No info level log message that an obsolete file was deleted?

A DAO method is already an atomic action: you only use explicit transaction when there are multiple DAO method calls.

If the OSL exists this method must return true, not just if cleanup is successful. The iteration has to skip this StorageLocation if it is obsolete. I would have this method throw if you cannot determine whether the StorageLocation is obsolete or not (fail to query). So: the `obsoleteStorageLocationDAO.get` has to be outside the `canTakeAction` so the return value is correct during a dryrun.

In this case, if the `delete(storageMetadata)` fails, you don't need to fail (throw IllegalStateException); the whole point of an OSL is that a failure to delete from storage doesn't have to interrupt activity and can be left to another try later since the OSL still exists. Whether a failure to delete is a reason to throw depends on what the failure is... I would probably *not* throw here as a result of failure to delete the obsolete file.

The log message right after obsoleteStorageLocationDAO.delete should be at info level and must be specific that this is an "obsolete" -- I think what is there on line ~705 would say that, but relying on toString() probably makes it uglier and more obtuse to operators than necessary. If it says "deleted ObsoleteStorageLocation[...]" that's really only saying an inventory object they don't know about  - because it is not in the model - was deleted from the database. The info message should be more carefully constructed to say that an obsolete file was removed from the system.

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
```diff
@@ -611,6 +617,10 @@ public void updateArtifact(final Artifact artifact, final StorageMetadata storag
             @Override
             public boolean hasNext() {
                 if (storageMetadataIterator.hasNext()) {
+                    storageMetadata = storageMetadataIterator.next();
```
pdowler, line 620: Calling code can call hasNext() multiple times for no good reason, so you cannot put the "advance" (call to storageMetadataIterator.next()) in here. I usually implement an advance() method that sets the "next item" field, so hasNext() is simply 
```
return storageMetadata != null;
```
Then you have to call advance() in the ctor and in next(), eg
```
public StorageMetadata next() {
   StorageMetadata ret = storageMetadata;
   advance();
   return ret;
}
```
And then advance() does the checking and skipping but only to find the first valid return object (ctor) and when caller actually advances (next).

pdowler (changes requested):

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
jburke-cadc: Now I remember using an advance() method in an iterator. Moved the MetadataStorage iterator method to a class, so advance() can be called in the class constructor. And a convenience method to get the class that can be overridden in the unit tests.

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
jburke-cadc: Updated the logging, removed the transactions, and the return value is based on the presence of an ObsoleteStorageLocation for the file.

pdowler: Emphasis added top previous comment. The isObsoleteStorageLocation method is still wrong -- the return value depends on whether or not this is a "dryrun". That means dryrun will tell the operator one thing and then normal mode will do something else

pdowler (changes requested):

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
jburke-cadc: Not seeing that in the code. `isObsoleteStorageLocation()`
1 - gets the OSL, which could be null if not found.
2 - if an OSL was found and it's not a dry run, delete file and OSL.
3 - if an OSL was found in 1 return true, else return false. 
The return value is based only on whether an OSL was found in 1.

pdowler (changes requested):
isObsoleteStorageLocation is still wrong. I highlighted a section in bold from a previous comment that pertains to this.

jburke-cadc:
I'm not seeing where isObsoleteStorageLocation is wrong. It is:
1 - getting the OSL for the StorageMetadata; the OSL is null if not found.
2 - if the OSL is not null and it is not a dry run, delete the file and the OSL
3 - return whether an OSL was found or not, based on the OSL returned in 1.
2 does not change what is returned in 3.

pdowler:
If it is a dryrun (canTakeAction() returns false) then isObsoleteStorageLocation always returns false. That is wrong. It has to return true if a matching OSL is found and false otherwise. It must fail (throw) if it cannot query to determine this fact.

isObsoleteStorageLocation could fail or just log an error and continue if it fails to perform the cleanup... I haven't thought about the pros and cons. For now, probably should fail as a log full of error messages isn't useful (after the first one) and failing means we'll fix errors rather than get used to them.

jburke-cadc:
My first thought is we can't be looking at the same code, and indeed we are not. I didn't push the last commit with the isObsoleteStorageLocation changes. Sorry for wasting your time with this. Pushed the last commit.

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
```diff
@@ -588,82 +591,139 @@ public void updateArtifact(final Artifact artifact, final StorageMetadata storag
...
+            throw new IllegalStateException(
+                String.format("query failed for ObsoleteStorageLocation %s for file %s",
+                              storageMetadata.getStorageLocation().getStorageID().toASCIIString(),
+                              storageMetadata.artifactURI.toASCIIString()), e);
+        }
+        if (obsoleteStorageLocation != null && canTakeAction()) {
+            try {
+                delete(storageMetadata);
+                this.obsoleteStorageLocationDAO.delete(obsoleteStorageLocation.getID());
+                LOGGER.info(String.format("deleted obsolete file %s at %s",
```
pdowler, line 645: The log message is inside the canTakeAction() so nothing is reported in reportOnly mode; that doesn't seem useful. An operator will run this with reportOnly=true, see nothing in the output, then run it for real and possibly see messages about files being deleted. Surprise!!

I don't know about the other uses of canTakeAction() vs logging, but if they are like this they suck. In general, you want the info log message before the attempted action, so "delete obsolete file ..." (not past tense). Further, since this message says it is deleting artifactURI it is very misleading: **artifactURI is not obsolete** and probably still exists (because an OSL is created when an artifact is overwritten as well as deleted). So something like this would be correct:
```
delete obsolete object: {storageLocation} (was: {artifactURI})
```
And you need to log the whole StorageLocation, not just the storageID - yeah, the default toString() is OK for now since it isn't obvious how to write storageBucket and storageID clearly without an unambiguous separator.

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
```diff
@@ -588,82 +591,139 @@ public void updateArtifact(final Artifact artifact, final StorageMetadata storag
...
+                                          storageMetadata.artifactURI.toASCIIString(),
+                                          obsoleteStorageLocation.getLocation().getStorageID().toASCIIString()));
+            } catch (Exception e) {
+                LOGGER.error(String.format("delete failed for obsolete file %s at %s",
```
pdowler, line 649: Aside from the same issue about correctness as above, this log message assumes that the `delete(storageMetadata)` was the action that failed, which might not be true. 
```
failed to cleanup obsolete {storageLocation}
```
More importantly, look at the now-emphasized text in my previous review comment concerning error handling.

pdowler (changes requested):

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
jburke-cadc: Most tantor methods that do work start with an `if (canTakeAction)` and there is no logging outside of that `if` so yeah running in dryrun mode is not going to tell the operator much.

jburke-cadc: Updated the logging to always log the intent in isObsoleteStorageLocation().

tantar/src/main/java/org/opencadc/tantar/BucketValidator.java
jburke-cadc: Updated to throw an exception if the cleanup fails.

pdowler (approved):

PR #271, opened 2021-02-24
CADC-9137 update fenwick to use cadc-inventory-util ArtifactSelector classes

jburke-cadc:


at88mph (approved):

pdowler (changes requested):
It is true that the ArtifactSelector is an interface and we generally configure plugins with 
```
{interface} = {impl}
```
However, I did not anticipate that this would impact configuration (users aka ops) like this and make this config item look even more developer oriented. I think we need to rethink this and only use {interface}={impl} oplugin style where it really matters: plugins with considerable code to adapt to external system (SQLGenerator and StorageAdapter).

So, what do you think about making the fenwick config item:
```
org,pencadc.fenwick.artifactSelector = all | filter
```
(and something equivalent in ratik). Note lower case so the config key doesn't look like a class name, just a property of fenwick (ratik).

Of course, the code in fenwick and ratik will instantiate known classes in cadc-inventory-util. 

Can you implement the necessary config parsing change in fenwick and ratik for this one config item and see if that looks better? If/when it does, we'll probably want to make the same kind of changes in a few other places, but that can be a separate ticket.

What I realised seeing this: refactoring to make code more maintainable should not result in a config changes that ops have to deal with, so using {interface}={impl} should only be used when we really need dynamic plugin loading and not just to avoid writing some config parsing.

jburke-cadc:
That sounds reasonable. Less stress on ops. They can choose and set the property entirely within the config file, without having to resort to looking through source code and then not quite sure if they've gotten it right or not.
I'll update fenwick & ratik.

pdowler (approved):

PR #275, opened 2021-03-23
CADC-9366 - only members of configured groups are allowed to query luskan

jburke-cadc:
Updated the `QueryRunnerImpl` in Luskan to override the `QueryRunner` run() method. run() in `QueryRunnerImpl` first calls checkPermission() to see if the caller is authorized, throwing an AccessControlException if the user is not authorized. If the user is authorized calls the `QueryRunner` run() method to do the query.
checkPermission() gets a list of groups the calling user is a member of. Then gets the list of a groups allowed to query luskan from the `org.opencadc.luskan.allowGroup` property in `luskan.properties`. Lastly compares the users groups with the allowed groups looking for a match.

### DIFF
--- a/luskan/src/intTest/java/org/opencadc/luskan/LuskanAsyncErrorTest.java
+++ b/luskan/src/intTest/java/org/opencadc/luskan/LuskanAsyncErrorTest.java
@@ -96,7 +96,7 @@ public class LuskanAsyncErrorTest extends TapAsyncErrorTest
     public LuskanAsyncErrorTest()
     { 
         super(Constants.RESOURCE_ID);
-        File testCertFile = FileUtil.getFileFromResource("x509_CADCAuthtest1.pem", LuskanSyncQueryTest.class);
+        File testCertFile = FileUtil.getFileFromResource("luskan-test-auth.pem", LuskanSyncQueryTest.class);
         setSubject(SSLUtil.createSubject(testCertFile));
         // re-use SyncResultTest files
         File testFile = FileUtil.getFileFromResource("AsyncErrorTest.foo.properties", LuskanSyncQueryTest.class);
--- a/luskan/src/intTest/java/org/opencadc/luskan/LuskanAsyncErrorTest.java
+++ b/luskan/src/intTest/java/org/opencadc/luskan/LuskanAsyncErrorTest.java
@@ -96,7 +96,7 @@ public class LuskanAsyncErrorTest extends TapAsyncErrorTest
     public LuskanAsyncErrorTest()
     { 
         super(Constants.RESOURCE_ID);
-        File testCertFile = FileUtil.getFileFromResource("x509_CADCAnontest1.pem", LuskanSyncQueryTest.class);
+        File testCertFile = FileUtil.getFileFromResource("x509_CADCAuthtest1.pem", LuskanSyncQueryTest.class);
         setSubject(SSLUtil.createSubject(testCertFile));
         // re-use SyncResultTest files
         File testFile = FileUtil.getFileFromResource("AsyncErrorTest.foo.properties", LuskanSyncQueryTest.class);
--- a/luskan/src/main/java/org/opencadc/luskan/LuskanConfig.java
+++ b/luskan/src/main/java/org/opencadc/luskan/LuskanConfig.java
@@ -78,7 +78,9 @@ import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
 import org.apache.log4j.Logger;
+import org.opencadc.gms.GroupURI;
 
 public class LuskanConfig {
     private static final Logger log = Logger.getLogger(LuskanConfig.class);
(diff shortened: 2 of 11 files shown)


luskan/src/main/java/org/opencadc/luskan/QueryRunnerImpl.java
```diff
@@ -103,4 +114,48 @@ protected DataSource getQueryDataSource() throws Exception {
         log.debug("Data Source name: jdbc/tapuser");
         return DBUtil.findJNDIDataSource("jdbc/tapuser");
     }
+
+    /**
+     *
+     */
+    @Override
+    public void run() {
```
pdowler, line 122: This is not the most upstream place to block access. In TAP, all jobs are uws jobs (sync and async) so the right place to check for global access is in JobPersistence.put(Job) - check auth before job creation means you get "permission denied" on the original job create and don't put a job in the database, rather than only getting it when running a job (where, if you are lucky this throw will give the right error from an async job but I doubt it).

luskan/src/main/java/org/opencadc/luskan/LuskanConfig.java
```diff
@@ -169,6 +172,21 @@ public static MultiValuedProperties getConfig() {
         }
         sb.append("\n");
 
+        List<String> allowedGroups = props.getProperty(ALLOWED_GROUP);
+        if (allowedGroups.isEmpty()) {
+            sb.append("MISSING");
```
pdowler, line 177: Since this is effectively required, `ok = false`??

luskan/src/main/java/org/opencadc/luskan/LuskanConfig.java
```diff
@@ -169,6 +172,21 @@ public static MultiValuedProperties getConfig() {
...
+        } else {
+            for (String allowedGroup : allowedGroups) {
+                sb.append("\n\t").append(ALLOWED_GROUP).append(" - ").append(allowedGroup);
+                try {
+                    new GroupURI(URI.create(allowedGroup));
+                    sb.append(" OK");
+                } catch (IllegalArgumentException e) {
+                    sb.append(" INVALID GroupURI");
```
pdowler, line 185: `ok = false`??

luskan/src/intTest/java/org/opencadc/luskan/LuskanAsyncErrorTest.java
```diff
@@ -96,7 +96,7 @@
     public LuskanAsyncErrorTest()
     { 
         super(Constants.RESOURCE_ID);
-        File testCertFile = FileUtil.getFileFromResource("x509_CADCAnontest1.pem", LuskanSyncQueryTest.class);
+        File testCertFile = FileUtil.getFileFromResource("x509_CADCAuthtest1.pem", LuskanSyncQueryTest.class);
```
pdowler, line 99: intTest/README.md above mentions some generic cert names  but they are not used in the test setup (here and other places). That's just necessary to run other tests.

There should be an AuthQueryTest that verifies:
1. anon query fails as expected
2. query by luskan-test-noauth fails as expected
3. for completeness, the same query by luskan-test-auth succeeds

pdowler (changes requested):

luskan/src/main/java/org/opencadc/luskan/QueryRunnerImpl.java
jburke-cadc: Added an AuthJobPersistence that extends PostgresJobPersistence and does the auth check before putting the job.

luskan/src/main/java/org/opencadc/luskan/LuskanConfig.java
jburke-cadc: Yes `ok=false` is missing in a couple of places.

luskan/src/main/java/org/opencadc/luskan/LuskanConfig.java
jburke-cadc: Added.

luskan/src/intTest/java/org/opencadc/luskan/LuskanAsyncErrorTest.java
jburke-cadc: Updated some int-tests to use the generic cert names and renamed the auth checking int-test to AuthQueryTest.

luskan/src/main/java/org/opencadc/luskan/QueryRunnerImpl.java
```diff
@@ -87,20 +86,17 @@
     public QueryRunnerImpl() {
     }
 
-    @Override
-    protected DataSource getUploadDataSource()
-            throws Exception {
+    @Override protected DataSource getUploadDataSource() throws Exception {
```
pdowler, line 89: Please revert this file to previous state - there were no changes needed but now the annotation is messed up

pdowler (changes requested):

luskan/src/main/java/org/opencadc/luskan/QueryRunnerImpl.java
jburke-cadc: I didn't change those methods so the IDE must've done something odd, although can't say I've seen it make formatting changes like that before. Fixed.

pdowler (approved):

PR #277, opened 2021-03-30
support init of buckets in SwiftStorageAdapter

pdowler:
create buckets  on init
store immutable config as bucket attrs in storage
fix intTest init to override some config from user config file

### DIFF
--- a/cadc-storage-adapter-swift/src/intTest/java/org/opencadc/inventory/storage/swift/SwiftByteRangeTest.java
+++ b/cadc-storage-adapter-swift/src/intTest/java/org/opencadc/inventory/storage/swift/SwiftByteRangeTest.java
@@ -94,7 +94,7 @@ public class SwiftByteRangeTest extends StorageAdapterByteRangeTest {
     final SwiftStorageAdapter swiftAdapter;
     
     public SwiftByteRangeTest() throws Exception {
-        super(new SwiftStorageAdapter(true, System.getProperty("user.name") + "-SwiftByteRangeTest", 3, false)); // single-bucket adapter for this
+        super(new SwiftStorageAdapter(true, System.getProperty("user.name") + "-single-test", 3, false)); // single-bucket adapter for this
         this.swiftAdapter = (SwiftStorageAdapter) super.adapter;
     }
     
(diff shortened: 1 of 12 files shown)


cadc-storage-adapter-swift/src/main/java/org/opencadc/inventory/storage/swift/SwiftStorageAdapter.java
```diff
@@ -259,23 +284,72 @@ public SwiftStorageAdapter() throws InvalidConfigException, StorageEngageExcepti
...
+                    }
+                } else {
+                    log.info("exists: " + bucketName);
+                }
             }
         }
+        
+        // init succeeeded: commit
+        Map<String,Object> meta = new TreeMap<>();
+        meta.put(VERSION_ATTR, "0.5.0");
```
andamian, line 348: What is this and how it's maintained?

cadc-storage-adapter-swift/src/intTest/java/org/opencadc/inventory/storage/swift/SwiftByteRangeTest.java
```diff
@@ -109,4 +109,9 @@ public void cleanupBefore() throws Exception {
         }
         log.info("cleanupBefore: DONE");        
     }
+    
+    @Test
+    public void testCleanupOnly() {
+        log.info("testCleanupOnly: no-op");
```
andamian, line 115: What is this for?

andamian (dismissed):

cadc-storage-adapter-swift/src/main/java/org/opencadc/inventory/storage/swift/SwiftStorageAdapter.java
pdowler: This is the version of the library at the point where the "schema" of the storage layout is created. If we have to change the schema at some point, then we'll have to run some code to upgrade from this version to the next version and then assign the new version attr. It's the same concept I use for database schema updates (via ModelVersion table). 

I don't know if we'll ever change the "storage schema" but this at least says which version of the library first required this version.

cadc-storage-adapter-swift/src/intTest/java/org/opencadc/inventory/storage/swift/SwiftByteRangeTest.java
pdowler: A no-op task allows one to run just this test method so you get the ctor (init) and cleanup code and no other tests. It's handy if you are having trouble with init (connection, auth, the init itself, or the cleanupBefore code).

PR #280, opened 2021-04-08
Cadc 9182 Add support to change minoc availability state

yeunga:
Added setAppName(), setState(), getState().


minoc/src/main/java/org/opencadc/minoc/ServiceAvailability.java
```diff
@@ -185,7 +196,27 @@ public AvailabilityStatus getStatus() {
...
+        if (RestAction.STATE_OFFLINE.equalsIgnoreCase(state)) {
+            System.setProperty(key, RestAction.STATE_OFFLINE);
+        } else if (RestAction.STATE_READ_ONLY.equalsIgnoreCase(state)) {
+            System.setProperty(key, RestAction.STATE_READ_ONLY);
+        } else if (RestAction.STATE_READ_WRITE.equalsIgnoreCase(state)) {
+            System.setProperty(key, RestAction.STATE_READ_WRITE);
+        } else {
+            throw new IllegalArgumentException("invalid state: " + state
+                + " expected: " + RestAction.STATE_READ_WRITE + "|" + RestAction.STATE_OFFLINE);
+        }
```
jburke-cadc, line 209: STATE_READ_ONLY missing in the IllegalArgumentException expected message.

jburke-cadc (approved):

minoc/src/main/java/org/opencadc/minoc/ServiceAvailability.java
yeunga: Fixed. Good catch.

minoc/src/intTest/java/org/opencadc/minoc/SetStateTest.java
```diff
@@ -0,0 +1,397 @@
...
+    }
+    
+    public SetStateTest() {
+        super();
+    }
+    
+    @Test
+    public void testSetOffline() throws Exception {
+        try {
+            // set minoc state to 'Offline'
```
pdowler, line 127: Test needs to confirm the initial state before testing that changes work; if it isn't in the right state to start (because another test left in the wrong state) then set it and confirm.

this comment applies to other test methods as well.

minoc/src/intTest/java/org/opencadc/minoc/SetStateTest.java
```diff
@@ -0,0 +1,397 @@
...
+            URL artifactURL = new URL(filesURL + "/" + artifactURI.toString());
+
+            String content = "abcdefghijklmnopqrstuvwxyz";
+            String encoding = "test-encoding";
+            String type = "text/plain";
+            byte[] data = content.getBytes();
+
+            // put: no length or checksum
+            InputStream in = new ByteArrayInputStream(data);
+            HttpUpload put = new HttpUpload(in, artifactURL);
```
pdowler, line 142: Please name the file with the name of the test method rather than just file.txt -- that way if it does get stored and left lying around it's easy to figure out which test created it.

None of this metadata is required so it's just extra code to look through and introduces ways for the test to fail for non-important reasons. 

Also applicable to other test methods.

minoc/src/intTest/java/org/opencadc/minoc/SetStateTest.java
```diff
@@ -0,0 +1,397 @@
...
+            URL artifactURL = new URL(filesURL + "/" + artifactURI.toString());
+
+            String content = "abcdefghijklmnopqrstuvwxyz";
+            String encoding = "test-encoding";
+            String type = "text/plain";
+            byte[] data = content.getBytes();
+
+            // put a file so that we can test get: no length or checksum
+            InputStream in = new ByteArrayInputStream(data);
+            HttpUpload put = new HttpUpload(in, artifactURL);
```
pdowler, line 216: Since this put is intended to be successful, the one header that should always be set is content-length (it may effect server side storage strategy in future). All the others are just noise as noted above.

minoc/src/intTest/java/org/opencadc/minoc/SetStateTest.java
```diff
@@ -0,0 +1,397 @@
...
+
+            Subject.doAs(userSubject, new RunnableAction(put));
+            Assert.assertNotNull("put to minoc in offline state should have failed", put.getThrowable());
+            String exMsg = put.getThrowable().getMessage();
+            Assert.assertTrue("incorrect exception message from put", exMsg.contains("System is offline"));
+
+            // get
+            OutputStream out = new ByteArrayOutputStream();
+            HttpGet get = new HttpGet(artifactURL, out);
+            Subject.doAs(userSubject, new RunnableAction(get));
```
pdowler, line 155: before asserts, it is useful to log (info) the response code and throwable... helps when figuring out why a test stopped passing. 

There should be an assert for the expected response code (before checking the throwable because if the code is wrong then something else happened and wrong code is more important). Once you add the above log message you can figure out what code you get now, add that assert, and then we'll decide if that's actually right.

pdowler (changes requested):

minoc/src/intTest/java/org/opencadc/minoc/SetStateTest.java
yeunga: Fixed.

minoc/src/intTest/java/org/opencadc/minoc/SetStateTest.java
yeunga: Removed checks on the metadata.

pdowler: I meant do not set the metadata in the headers at all (setRequestProperty and setDigest below).

pdowler (changes requested):
Not fixed. The initial state is always ReadWrite mode; must verify that the test starts in that mode.
 
The test is to change to Offline mode and verify 
* that changing to Offline mode actually works: get the availability and verify what it says
* changes behaviour of the service: try all the ops and verify that they succeed or fail (all fail here)
End of test should change back to ReadWrite mode and verify.

Same logic for ReadOnlyMode; different expectations w.r.t head and get ops

There should probably be a test to verify that one can transition between Offline and ReadOnly correctly; since the main tests verify service behaviour, the Offline-to-ReadOnly and reverse only need to get the availability and trust it to be correct.

minoc/src/intTest/java/org/opencadc/minoc/SetStateTest.java
yeunga: Removed the metadata, except for content-length in a put. Renamed file.txt.

minoc/src/intTest/java/org/opencadc/minoc/SetStateTest.java
yeunga: Added content-length to put.

minoc/src/intTest/java/org/opencadc/minoc/SetStateTest.java
yeunga: Added log (info) of response code and throwable. Added an assert for the expected response code.

pdowler (approved):

PR #286, opened 2021-04-13
CADC-9080 update fenwick to retry on errors

jburke-cadc:


fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
```diff
@@ -197,9 +224,35 @@ public void run() {
...
+
+                // successful run, reset retry values
+                retry = false;
+                retryCount = 1;
+                sleepSeconds = INITIAL_SLEEP_TIMEOUT;
             } catch (PrivilegedActionException privilegedActionException) {
-                final Exception exception = privilegedActionException.getException();
-                throw new IllegalStateException(exception.getMessage(), exception);
+                // Get the check exceptions in PrivilegedActionException thrown by doit()
+                final Exception cause = privilegedActionException.getException();
```
pdowler, line 234: Ugh. This PrivilegedActionException thing is a pain because the resulting conditional is really error prone.

First, `doit()` should never be throwing NoSuchAlgorithmException: track down where that comes from, catch it as locally as possible, and wrap it in a `new RuntimeException("BUG: failed to get instance of ?", ..., ex)`.

Second issue you did not seem to recognise: we use RuntimeException for local bugs and unexpected failures (as above) all over the place, and the cadc-rest library translates a RuntimeException into a 500, so the HttpTransfer (client and subclasses) translates a 500 back into a RuntimeException... that's seems fine except you cannot tell the difference between a local and remote RuntimeException. So, the first thing to do is probably go into cadc-util and add a new Exception: `RemoteServiceException extends RuntimeException` and use that for a 500 response in `HttpTransfer.checkErrors`. Without that you will not be able to deal with the important case (some service responding with a  500 because it's broken).

And then there is this PrivilegedExceptionAction bit that completely messes up exception handling. The current code organisation means you can't use catch (multi-catch) to handle exceptions so I think you have to rethink where the Subject.doAs(...) lives. It's in the current spot because it was the easy way to re-read the proxy cert file before each query rather... I think all this error handling has to be inside there so you can use normal exception handling. Remember that Subject.doAs(...) can return a value (the `PrivilegedExceptionAction<Void>` just says it doesn't, but the type arg could be Boolean or String or whatever class you need to convey info from inside... or you could just handle all retry-able exceptions inside and any throw is fatal. 

Side note: it took me quite a bit of work to modify the ca.nrc.cadc.net package to have a prepare() method that did all the work (except the transfer) and throw declared exceptions in the normal way, but it is a comparable thing. The old style of `run()` and then `getThrowable()` was fine if you were going to just print the message, but horrible if you wanted to actually write code to react to different failures. The use or misplacing of a Subject.doAs(...) wrecks havoc on error handling.... so this will take some thought to come up with a clean and maintainable solution.

pdowler (changes requested):

fenwick/README.md
```diff
@@ -45,6 +49,11 @@ remote. If it is `filter` (was: `org.opencadc.fenwick.IncludeArtifacts`) then fe
 from remote (see artifact-filter.sql below). A global inventory and a storage site that should get all Artitacts 
 (files) would run with `all`. Specialised instances that want to select a subset of all files would use the explicit filtering. 
 
+`retriesTimeout` is the number of seconds fenwick will attempt to retry processing after encountering an error.
+If fenwick encounters a non-fatal error, it will sleep for an initial timeout value, and run again. 
+If a subsequent run encounters an error, it will double the previous timeout value, and sleep again before 
+trying another run. When the total timeouts exceed `retriesTimeout` fenwick will stop processing and exit.
+
```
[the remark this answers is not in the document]
jburke-cadc: Read the issue as: keep retrying until a timeout value is reached, at which point the system is broken, so exit until it is fixed and fenwick can be restarted again. Updated so the maxRetryInterval is the maximum timeout value, and fenwick will keep retrying at that value until it completes a successful run.

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
```diff
@@ -185,7 +194,25 @@ public InventoryHarvester(Map<String, Object> daoConfig, URI resourceID, Artifac
...
+                    throw new RuntimeException(String.format("Exiting, retries timeout of %s seconds exceeded",
+                                                             this.retriesTimeout));
+                }
+                try {
+                    Thread.sleep(sleepSeconds * 1000L);
+                } catch (InterruptedException ex) {
+                    throw new RuntimeException("thread sleep interrupted");
+                }
+                retryCount++;
+                sleepSeconds += sleepSeconds;
```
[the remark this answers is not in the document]
jburke-cadc: I don't think I've ever used *= before. It's easier to read, which would make it clearer. Updated.

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
jburke-cadc: Removed the throws declarations for NoSuchAlgorithmException, now catch NoSuchAlgorithmException where it is thrown and then throw a RuntimeException.

Added a RemoteServiceException class to the cadc-util net classes, updated HttpTransfer to throw it for a HTTP 500 error.

Updated the Subject.doAs to use a PrivilegedAction and catch the checked exceptions inside the run() method. The action returns a boolean indicating whether an exception was thrown that should result in a retry.

Fatal runtime exceptions are caught in the try around the Subject.doAs.

You mentioned moving the Subject.doAs out of the loop. You could put the .doAs into a method and call the method, but I think it's better to keep all the exception handling and retry/fail logic in the same place. Easier to see and understand what it is doing. Open to being wrong though :-)

fenwick/README.md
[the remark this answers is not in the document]
pdowler: the word timeout (or timout) is not applicable; it is not a timeout - a timeout implies that something stops happening which is not the case here

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
```diff
@@ -185,43 +194,108 @@ public InventoryHarvester(Map<String, Object> daoConfig, URI resourceID, Artifac
...
         while (true) {
             try {
                 final Subject subject = SSLUtil.createSubject(new File(CERTIFICATE_FILE_LOCATION));
-                Subject.doAs(subject, (PrivilegedExceptionAction<Void>) () -> {
-                    doit();
-                    return null;
+                int retries = retryCount;
+                int timeout = sleepSeconds;
+
+                retry = Subject.doAs(subject, (PrivilegedAction<Boolean>) () -> {
```
pdowler, line 207: The docs for PrivilegedAction say that run() does not throw checked exceptions, but I don't think checked vs unchecked is really the same as "something I can retry" vs "fatal". You will have to catch all checked exceptions inside run() even if some of those should be fatal. So the outer catch for fatal can only be unchecked (RuntimeException).

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
```diff
@@ -185,43 +194,108 @@ public InventoryHarvester(Map<String, Object> daoConfig, URI resourceID, Artifac
...
+                    boolean isRetry = true;
+                    try {
+                        doit();
+                        isRetry = false;
+                        // catch exceptions resulting in a retry
+                    } catch (ResourceNotFoundException ex) {
```
pdowler, line 213: Why would you create new error messages with the http status code (which will eventually be wrong as code elsewhere evolves) when you could just do a multi-catch of all non-fatal exceptions and use the exception name in the error message. Even ex.toString() would be more or less OK.

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
```diff
@@ -185,43 +194,108 @@ public InventoryHarvester(Map<String, Object> daoConfig, URI resourceID, Artifac
...
+                    } catch (RemoteServiceException ex) {
+                        // 500 internal service error, retry.
+                        logRetry(retries, timeout, "HTTP 500 runtime error", ex.getMessage());
+                    } catch (TransientException ex) {
+                        //  503 transient, retry.
+                        logRetry(retries, timeout, "HTTP 503 transient", ex.getMessage());
+                    } catch (IOException ex) {
+                        // IO error talking to service, retry.
+                        logRetry(retries, timeout, "IO error", ex.getMessage());
+                    } catch (IllegalArgumentException ex) {
```
pdowler, line 234: Practical issue: the IllegalArgumentException below (part of the streaming fail hack) is ambiguous. The hack is catching this from the client side RowMapper that gets an incomplete row, but at end there is a comment that it is a service 400 error (because cadc-rest and cadc-util map IllegalArgumentException -> 400 -> IllegalArgumentException: so, you cannot tell if that is the service failing or the client failing. So, this exception should not be in the "keep trying" list at all. 

I think we need to retain detecting the truncated streaming but that probably needs to be done in the TapClient or RowMapper impls. I feel like the TapClient could be checking row data before handing off to RowMapper and it knows if end-of-stream was reached so should be able to detect end-of-stream and incorrect row data better than this. Need to discuss.

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
```diff
@@ -185,43 +194,108 @@ public InventoryHarvester(Map<String, Object> daoConfig, URI resourceID, Artifac
...
+                    // TODO: dynamic depending on how rapidly the remote content is changing
+                    // ... this value and the reprocess-last-N-seconds should be related
+                    Thread.sleep(DEFAULT_SLEEP_TIMEOUT * 1000L);
                 }
             } catch (InterruptedException ex) {
-                throw new RuntimeException("interrupted", ex);
+                // Thread interrupted, fail.
+                logExit("Thread interrupted", ex.getMessage());
+                throw new RuntimeException(ex.getMessage(), ex);
+            } catch (NotAuthenticatedException ex) {
```
pdowler, line 287: The InterruptedException should not be caught and wrapped. Just let it fly.

The NotAuthenticatedException from HttpTransfer if it thinks the proxy cert in the subject was rejected by the server (eg expired), so this is actually a failure that could start working if ops updates the cert in the config dir. So technically recoverable without restarting fenwick.

The AccessControlException from HttpTransfer will be due to permissions at the remote luskan and could start working is ops fixes some other config (usually baldur or luskan) or replaces the proxy cert with the right one. So technically recoverable without restarting fenwick.

pdowler (changes requested):

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
```diff
@@ -185,43 +194,108 @@ public InventoryHarvester(Map<String, Object> daoConfig, URI resourceID, Artifac
...
-                    log.error("\n*******\n");
-                } else if (message.startsWith("invalid checksum URI:")) {
-                    log.error("\n\n*******\n");
-                    log.error("Caught IllegalArgumentException - " + message + " (" + ++errorCount + ")");
-                    log.error("Ignoring error as presumed to be a dropped connection before fully reading the stream.");
-                    log.error("CAUTION! - This could actually be a bad MD5 checksum! Logging this to provide an audit.");
-                    log.error("\n*******\n");
+                if (retry) {
+                    Thread.sleep(sleepSeconds * 1000L);
+                    retryCount++;
```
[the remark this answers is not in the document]
jburke-cadc: Cleaned up the sleep timeout calculation and use a single Thread.sleep.

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
jburke-cadc: Moved all exception checking into the run(), except for a RuntimeException wrapping the Subject.doAs.

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
jburke-cadc: Used a multi-catch for the run() exceptions that result in a retry, and use the exception message for the retry logging.

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
jburke-cadc: Updated the TapClient to let it throw an IndexOutOfBoundsException when it tries to read a truncated stream.

fenwick/src/main/java/org/opencadc/fenwick/InventoryHarvester.java
jburke-cadc: The InterruptedException is a checked exception in the run method that has to be caught.

I asked myself if a certificate expired, what would ops notice first: an error message in a log, or the fact that fenwick had stopped running. 

If it's expected that ops will detect and fix cert issues, then should retry on cert exceptions, updated.

pdowler (dismissed):

PR #293, opened 2021-04-21
Cadc 9509

andamian:
Read-only sites at the end of negotiated URL list


pdowler (changes requested):
Isn't `protos.add(protos.size() - 1, p);` the same as the original `protos.add(p);` eg add to the end?

I'd probably invert the check and use
```
if (storageSite.getAllowWrite()) {
   // high priority for full read/write sites
   protos.add(0, p);
} else {
   protos.add(p);
}
```
Note that this is already inside an `if (storageSite.getAllowRead())` block (line 176).

Also note that there is sometimes another protocol added down at line ~214.

This is a quick hack and probably not worth writing a test for. When we try to use other logic to order the result URLs we'll have to restructure this, at which time it would be good to do that in a way that the ordering function could be tested with unit tests.

pdowler (approved):

PR #294, opened 2021-04-21
CADC-1166 CADC-8841 resolves the Artifact discrepancies, int-test fixes

jburke-cadc:


ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -168,7 +166,31 @@ protected void validateLocal(Artifact local)
         Artifact remote = getRemoteArtifact(local.getURI());
         if (remote != null) {
             log.info(String.format("delete: %s %s reason: local filter policy change", local.getID(), local.getURI()));
-            //TODO do action
+            try {
```
pdowler, line 169: Only one review comment on this particular block: ArtifactValidator lines 168-195. Please concentrate on fixing this one block before making similar changes in others I haven't looked at.

In general if you using an explicit transaction, you must have a try/catch/finally. Almost always: the startTransaction should be the first thing in the try and the commitTransaction should be the last thing in the try. There must be a catch(Exception) that does a rollback after any error, and there must be a finally that check if the transaction is open and rollback and report error very loudly. See minoc PutAction (near the bottom).

The log.info for this case (currently line 168) should be the line right before `artifactDAO.delete(...)` so it only gets printed if something actually happens.

The EntityNotFoundException from the lock call means that sometime between reading the Artifact from the database and now someone else deleted it. Since you only need the entity ID you do not need `get(local.getID())` again and (after the lock) local can never be null, so just proceed to do the actions and commit the transaction.

In the catch block, you may have to be more careful in calling rollback(). Some code re-org above that puts a new failure mode inside the try but before the start (or after the commit) will cause the rollback to fail and be really confusing. For example, in minoc PutAction: there is some code in the try block after the commit (legit in that case) that could throw, and then the rollback would fail. that code was moved and the catch block not updated to deal with a failure after the commit. So, if you can have commit be the last thing in the try, the catch can remain simple. You really need a catch(EntityNotFoundException) that does debug logging and rollback, and a general catch(Exception) that does error logging and rollback.

I also don't see what checking for a DeletedArtifactEvent does here. if you get an EntityNotFoundException from the lock attempt, it means the local artifact is (now) gone so the message is actually wrong. It's more like "I thought I saw X, but now it's gone so nevermind" :-)

And the finally block to detect open transactions!! Technically, the rollback calls should be in little try/catch of their own (the compiler lets you get away with it because spring framework uses RuntimeException for everything) because they can fail as well and it would be good to log.error such a fail. I might have to put a thrown checked exception in TransactionManager...

aside: if you have rollback in the try block, that's usually a sign that you are doing something wrong.; it's also usually a bad sign when the startTransaction() and the commitTransaction() are not in the same block (eg here they are at different indent levels with conditional, which is always dangerous)...  I wouldn't say always, but usually a bad sign.

pdowler (changes requested):

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
jburke-cadc: Reworked the transactions and error handling.

pdowler (changes requested):
Feedback on the transaction content included below.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -178,9 +210,35 @@ protected void validateLocal(Artifact local)
...
-                                   local.getID(), local.getURI()));
-            //TODO do action
+            try {
+                log.debug("starting transaction");
+                this.transactionManager.startTransaction();
+                log.debug("start txn: OK");
+
+                this.artifactDAO.lock(local);
+                log.info(String.format("delete: %s %s reason: found remote DeletedArtifactEvent",
+                                       local.getID(), local.getURI()));
```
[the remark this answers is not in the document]
jburke-cadc: Updated to persist the remote DeletedArtifactEvent locally.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -193,11 +251,46 @@ protected void validateLocal(Artifact local)
...
-                //TODO do action
-                return;
+                if (local.siteLocations.contains(remoteSiteLocation)) {
+                    try {
+                        log.debug("starting transaction");
+                        this.transactionManager.startTransaction();
+                        log.debug("start txn: OK");
+
+                        this.artifactDAO.lock(local);
+                        // if siteLocation's becomes empty removing the siteLocation, delete the artifact
```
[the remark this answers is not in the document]
jburke-cadc: Updated to get the current local Artifact after the lock, and check again if it contains the remote SiteLocation before removing it.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -274,9 +425,34 @@ protected void validateRemote(Artifact remote)
...
-                    log.info(String.format("add: %s to %s %s reason: remote SiteLocation missing",
-                                           remoteSiteLocation, remote.getID(), remote.getURI()));
-                    //TODO do action
+                    try {
+                        log.debug("starting transaction");
+                        this.transactionManager.startTransaction();
+                        log.debug("start txn: OK");
+
+                        this.artifactDAO.lock(local);
+                        log.info(String.format("add: %s to %s %s reason: remote SiteLocation missing",
```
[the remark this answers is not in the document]
jburke-cadc: Getting the Artifact after the lock and checking the conditions again.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -209,7 +302,45 @@ protected void validateLocal(Artifact local)
...
-            log.info(String.format("remove SiteLocation: %s %s reason: multiple", local.getID(), local.getURI()));
+            SiteLocation remoteSiteLocation = new SiteLocation(this.remoteSite.getID());
+            if (local.siteLocations.contains(remoteSiteLocation)) {
+                try {
+                    log.debug("starting transaction");
+                    this.transactionManager.startTransaction();
+                    log.debug("start txn: OK");
+
+                    this.artifactDAO.lock(local);
+                    // if siteLocation's becomes empty removing the siteLocation, delete the artifact
```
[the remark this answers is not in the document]
jburke-cadc: Same as above.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -322,9 +529,35 @@ protected void validateLocalAndRemote(Artifact local, Artifact remote) {
...
                 // action: put Artifact
-                log.info(String.format("resolve Artifact.metaChecksum mismatch: put remote %s %s "
-                                           + "reason: remote lastModified newer than local",
-                                       remote.getID(), remote.getURI()));
+                try {
+                    log.debug("starting transaction");
+                    this.transactionManager.startTransaction();
+                    log.debug("start txn: OK");
+
+                    this.artifactDAO.lock(local);
```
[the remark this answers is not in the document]
jburke-cadc: Same as above.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
```diff
@@ -432,22 +435,38 @@ String buildRemoteQuery(final String bucket)
         return query.toString();
     }
 
+    /**
+     * Method that can be overridden in tests to insert Artifact's after the iterator queries.
+     */
+    void testAction() { }
```
pdowler, line 441: Is this `testAction()` here so you can simulate the "artifact changed while iterating" that gets detected after acquiring a lock? It is admittedly hard to test that code handles race conditions (two different threads/processes are changing state and the actual sequence of events is a "race"). I haven't tried to do it before because there are so many unknowns and I'd normally rely on careful code review/analysis.... 

Methods like this (and in general odd/unnatural code to enable testing) make me nervous. Plus you can't really know if any action you take will or will not be visible in the iterator (eg you add a row to the db and it could show up in the iterator or not depending on all kinds of subtle things like row fetch size, query plan, db locking, etc).

I haven't looked at the test changes to make use of this, but I would expect this to be fragile at best, while the "lock && get" pattern in code is easy to look for. 

So, do you think this is really worth it? Are the tests going to hold up? Are the going to work and be tests easy to grok in six months?

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -322,9 +532,44 @@ protected void validateLocalAndRemote(Artifact local, Artifact remote) {
...
+                    Artifact current = this.artifactDAO.get(local.getID());
+                    if (!current.getMetaChecksum().equals(remote.getMetaChecksum())) {
+                        if (current.getLastModified().before(remote.getLastModified())) {
+                            log.info(String.format("resolve Artifact.metaChecksum mismatch: put remote %s %s "
+                                                       + "reason: remote lastModified newer than local",
+                                                   remote.getID(), remote.getURI()));
+                            this.artifactDAO.put(remote);
```
pdowler, line 547: Probably the final thing here:  the `put(remote)` has do more than just make the state equal to the remote state. If this is a global db, it has to "merge" remote changes in without losing any siteLocations and if it is a storage site it has to update metadata without losing the storageLocation. 

See lines 615-634 in fenwick InventoryHarvester: since the scenario here is mismatched checksums you are doing more or less the `else` part of that.

pdowler (changes requested):

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
jburke-cadc: Updated ArtifactValidator to preserve the local Artifacts site and storage locations.

ratik/src/main/java/org/opencadc/ratik/InventoryValidator.java
jburke-cadc: Yes, the `testAction()` method is solely for a single test case where after the local and remote iterators have been created an Artifact is put into the local site.
I agree it's a hack for a single test case, but it was the only way I could come up with to test that discrepancy.
Is it worth it? For a single test case probably not. Maybe if it was useful for multiple cases we could pretend not to see it.
Will they work in 6 months? Depends on if they are maintained or not. Easy to understand comes from documentation and comments in the tests. I think the docs explains the each test case well enough. More comments in the tests about what they are doing and why could be helpful down the road. I'll go through and add comments to the tests to explain them better.

PR #304, opened 2021-05-01
CADC-9537 Optimize performance using Protocol.securityMethod on Subject.doAs(...)

yeunga:


pdowler (changes requested):
The tests don't really prove that the code is doing the (subtle) correct thing because the transfer negotiation asks for and gets back multiple URLs and the target files are ultimately public and can be downloaded anonymously. I can't see any obvious way to test that by running an entire FileSyncJob, so just be aware that "working" doesn't mean "correct".

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
```diff
@@ -102,8 +104,9 @@
 public class FileSyncJobTest {
     private static final Logger log = Logger.getLogger(FileSyncJobTest.class);
     private static final String TEST_ROOT = "build/tmp/fsroot/critwallTests";
-    private static final String TEST_ARTIFACT_URI = "ad:IRIS/I212B2H0.fits";
-    private static final String TEST_RESOURCE_ID = "ivo://cadc.nrc.ca/data";
+    private static final String TEST_ARTIFACT_URI = "cadc:IRIS/I212B2H0.fits";
+    private static final String TEST_RESOURCE_ID = "ivo://cadc.nrc.ca/global/raven";
+    private static final String CERTIFICATE_FILE_LOCATION = System.getProperty("user.home") + "/.ssl/cadcproxy.pem";
```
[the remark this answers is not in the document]
yeunga: Done

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -309,102 +304,115 @@ private void doSync() {
...
+            FileContent content = new FileContent(out.toByteArray(), "text/xml");
+            log.warn("xml file content to be posted: " + transfer);
+    
+            log.debug("artifact path: " + artifact.getPath());
+            HttpPost post = new HttpPost(transferURL, content, true);
+            post.prepare();
+            log.debug("post prepare done");
+    
+            TransferReader reader = new TransferReader();
+            Transfer t = reader.read(post.getInputStream(), null);
```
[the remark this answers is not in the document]
yeunga: Thanks. Updated algorithm to return List<Protocol>.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -309,102 +304,115 @@ private void doSync() {
...
+        Subject currentSubject = new Subject();
+
+        // Also synchronized in FileSync.run()
+        synchronized (subject) {
+            currentSubject.getPrincipals().addAll(subject.getPrincipals());
+            currentSubject.getPublicCredentials().addAll(subject.getPublicCredentials());
+        }
+
+        Subject subj = AuthenticationUtil.getAnonSubject();
+        AuthMethod am = AuthenticationUtil.getAuthMethodFromCredentials(currentSubject);
```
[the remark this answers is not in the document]
yeunga: Changed to use anonSubject on anon SecurityMethod.

critwall/src/intTest/java/org/opencadc/critwall/FileSyncJobTest.java
```diff
@@ -117,9 +120,16 @@
 
     private ArtifactDAO dao = new ArtifactDAO();
     private Subject anonSubject = AuthenticationUtil.getAnonSubject();
+    private Subject certSubject = null;
 
     public FileSyncJobTest() throws Exception {
         try {
+            final File certificateFile = new File(CERTIFICATE_FILE_LOCATION);
+            this.certSubject = certificateFile.exists()
```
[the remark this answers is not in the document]
yeunga: Oops! Good catch. Fixed.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -309,102 +304,115 @@ private void doSync() {
...
-
         while (urlIterator.hasNext()) {
             URL u = urlIterator.next();
             log.debug("trying " + u);
-
             boolean postPrepare = false;
             try {
                 ByteArrayOutputStream dest = new ByteArrayOutputStream();
                 HttpGet get = new HttpGet(u, dest);
                 get.prepare();
```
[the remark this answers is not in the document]
yeunga: Refactored the code around get.prepare.

critwall/VERSION
```diff
@@ -1,4 +1,4 @@
 ## deployable containers have a semantic and build tag
 # semantic version tag: major.minor[.patch]
 # build version tag: timestamp
-TAGS="0.3-$(date --utc +"%Y%m%dT%H%M%S")"
+TAGS="0.4-$(date --utc +"%Y%m%dT%H%M%S")"
```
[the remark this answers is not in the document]
yeunga: Okay. Thanks. Reverted.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -306,105 +301,147 @@ private void doSync() {
...
             log.debug("trying " + u);
+            Subject subj = subject;
 
+            if (p.getSecurityMethod() == null || p.getSecurityMethod().equals(Standards.getSecurityMethod(AuthMethod.ANON))) {
+                subj = AuthenticationUtil.getAnonSubject();
+            }
+
+            try {
+                return Subject.doAs(subj, new SyncArtifact(u, a));
+            } catch (RuntimeException ex) {
```
pdowler, line 388: Using the message in a RuntimeException for application logic is really bad practice. 

The previous code used the different catch blocks to remove a url (now protocol) from the iterator and it was clear which exceptions were judged transient (don't remove) vs fatal for that endpoint (remove) vs fatal for the job (don't catch or rethrow). The SyncArtifact.run() declares a bunch of exceptions, so just throw them and let the caller sort it out as before.

`Subject.doAs(...)` does make it harder to do error handling but this is now worse than it was before: fragile and easily broken by text changes. It's possible that you moved too much of the code from the iterator loop into the SyncArtifact action: the chosen subject is only needed to call `get.prepare()` and after that A&A has succeeded and the InputStream can be read. that means a lot of error handling for back end failures has to also transition out of that Subject.doAs(...) ... 

Please re-think the error handling here. Ask (slack) if you want advice on how to reorganise this.

pdowler (changes requested):

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -306,105 +301,147 @@ private void doSync() {
...
-        Subject subject = AuthenticationUtil.getCurrentSubject();
-        AuthMethod am = AuthenticationUtil.getAuthMethodFromCredentials(subject);
-        log.debug("resource id: " + resource);
-        URL transferURL = regClient.getServiceURL(resource, Standards.SI_LOCATE, am);
-        if (transferURL == null) {
-            transferURL = regClient.getServiceURL(resource, Standards.VOSPACE_SYNC_21, am);
-        }
-        log.debug("certURL: " + transferURL);
+        TransientException, TransferParsingException, PrivilegedActionException {
+        Subject currentSubject = new Subject();
```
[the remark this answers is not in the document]
yeunga: Updated code to use a sharedSubject, a localSubject and an anonSubject.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
yeunga: Refactored SyncArtifact to handle all Exceptions in the caller. Currently the Java compiler complains about catching a couple of Exceptions in the caller that are not thrown in SyncArtifact. However those Exceptions are being thrown in SyncArtifact.run().

pdowler (changes requested):
There is a detail about Subject.doAs(...) that I now realise you probably don't know: the AccessControlContext is a stack and if you do a `Subject.doAs` inside a previous one, it changes the current subject and then restores the previous one when the `Subject.doAs` is complete.

Given that, the previous code organisation of having run() make a local copy of the shared subject and then run everything in a Subject.doAs essentially meant: do everything with this identity unless there is a specific reason to do something else. That encompassed transfer negotiation and *all* downloads. The change in behaviour was to do *some* downloads anonymously instead of using credentials (reason: because when a system is in a degraded state, authenticating might cause a fail when anon would work). But the important thing is that the only change was to "sometime* do another Subject.doAs(anon, ...) while within the default AccessControlContext.

I prefer the clarity and simplicity of run() the way it was. It's always possible that any code being executed might want to look at the subject and use the credentials. With your changes, lots of code now runs in the AccessControlContext of the thread pool instead of the job and specified subject, so it is also only right if none of that code ever needs a subject (now or in future).

I suggest that you make a copy of your FileSyncJob code for reference and revert the one in git to the original. Then write an nested action class to use to execute the HttpGet.prepare() method. Writing that action to create and return an HttpGet instance where some methods have been called, but not others, so the state is not clear from the api: that's poor style. I would leave the `new HttpGet` where it is now and then execute the prepare() method in a Subject.doAs. Should be minimal changes around line 390 (of the current code) to sometimes use an anonSubject.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -208,9 +210,8 @@ private void doSync() {
                     }
                     
                     // attempt to sync file
-                    log.debug("FileSyncJob.SYNC " + artifactLabel + " urls=" + urlList.size() + " attempts=" + retryCount);
+                    log.info("FileSyncJob.SYNC " + artifactLabel + " urls=" + urlList.size() + " attempts=" + retryCount);
```
pdowler, line 213: this message is supposed to be at debug level; the intent is to have 1 "start" and 1 "end" message and maybe we'll get rid of the "start" message in future; don't want any info messages in the loop.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -282,12 +280,12 @@ private void doSync() {
...
                     
                     if (!success) {
-                        log.info("FileSyncJob.SLEEP " + artifactID + " dt=" + RETRY_DELAY[retryCount]);
+                        log.info("FileSyncJob.SLEEP dt=" + RETRY_DELAY[retryCount]);
                         Thread.sleep(RETRY_DELAY[retryCount++]);
                     }
                 }
                 if (!success) {
-                    msg = " retryCount=" + retryCount + " download-attempts=" + numAttempts;
+                    msg = " all URLs tried, retryCount=" + retryCount;
```
pdowler, line 288: please remove the extra cruft and use the msg that was there; if/when we convert to some structured logging or ingest into a queryable system that free text is just in the way.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -354,37 +352,31 @@ private void doSync() {
...
             boolean postPrepare = false;
             try {
                 ByteArrayOutputStream dest = new ByteArrayOutputStream();
                 HttpGet get = new HttpGet(u, dest);
+                if (p.getSecurityMethod() == null || p.getSecurityMethod().equals(Standards.getSecurityMethod(AuthMethod.ANON))) {
+                    get = Subject.doAs(anonSubject, new NestedAction(u));
+                } else {
+                    get.prepare();
+                }
+
```
pdowler, line 379: Ack!!! Those timeouts *must* be set before you make the call - right after creating the HttpGet instance! They can't do any good after the call.

And I explicitly said to use the HttpGet instance created here, not create a new one in the action and return it: for the very reason that it makes this kind of mess where important state is not called or not known. Why create an instance (line 373) then immediately ignore it, pass the raw url into the action, and create a new one? 

And you still have another call to get.prepare() on line 382.

That's 4 bugs caused by one cut&paste!

pdowler (changes requested):

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
yeunga: Sorry about the careless errors. Fixed, I think.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
yeunga: Okay. Changed back to debug level.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
yeunga: Removed msg on failure.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -387,7 +370,12 @@ private StorageMetadata syncArtifact(Artifact a, List<URL> urls)
                 HttpGet get = new HttpGet(u, dest);
                 get.setConnectionTimeout(6000); // ms
                 get.setReadTimeout(60000);      // ms
-                get.prepare();
+                if (p.getSecurityMethod() == null || p.getSecurityMethod().equals(Standards.getSecurityMethod(AuthMethod.ANON))) {
+                    get = Subject.doAs(anonSubject, new NestedAction(get));
```
pdowler, line 374: still assigning to `get`

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -442,4 +430,22 @@ private StorageMetadata syncArtifact(Artifact a, List<URL> urls)
...
+        private HttpGet get;
+
+        // Hack: throw the exceptions so that Java compiler won't complain when the caller
+        //       tries to catch these exceptions.
+        NestedAction(HttpGet g) {
+            this.get = g;
+        }
+
+        @Override
+        public HttpGet run() throws ByteLimitExceededException, IllegalArgumentException,
```
pdowler, line 443: Still returning an HttpGet

pdowler (changes requested):

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
yeunga: Removed.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
yeunga: Removed.

pdowler (approved):

PR #336, opened 2021-07-12
Added META=true support.

at88mph:


minoc/src/main/java/org/opencadc/minoc/GetAction.java
```diff
@@ -127,14 +135,35 @@ public void doAction() throws Exception {
...
+                        for (final Cursor<String, HeaderCard> headerCardCursor = header.iterator();
+                             headerCardCursor.hasNext();) {
+                            final HeaderCard headerCard = headerCardCursor.next();
+                            bufferedWriter.write(headerCard.toString());
+                            bufferedWriter.newLine();
+                        }
+                    }
+                    bufferedWriter.flush();
+                } else {
+                    HeadAction.setHeaders(artifact, syncOutput);
```
pdowler, line 161: I don't like the use of multi-level logic here and especially having the primary action (download complete file) buried in the middle of it. There should be a clear "if no-params XXX else YYY" at the top level. The  complete download (and eventually content-range requests) will be in XXX and all operations will be in YYY. 

Once inside YYY, one can deal with conflicting params and throw appropriate errors. The proposed code would silently ignore META=true is there were also SUBS, and soon we'll have META vs SUBS vs POS/BAND/TIME/POL/++ to manage.

Conflicting param usage should throw an IllegalArgumentException.

Smaller issues: 
- the FitsOperations should not be instantiated unless it will be used
- the output format of ascii headers should be implemented in FitsOperations to follow the style of cutoutToStream... `fitsOperations.headersToStream(OutputStream)`??
- technically, the FitsOperations should only be attempted if the Artifact.contentType says the file is FITS. The two plausible types are `application/fits` and `image/fits` -- those constants could be declared in an upstream library, but I don't know which one at this point so maybe keep it here for now (we have mime-type constants all over the place right now and I would probably consolidate them in cadc-util witth the ContentType class, somehow)

pdowler (changes requested):

pdowler (approved):

PR #343, opened 2021-07-27
CADC-8668 raven checks storage site availability before returning URL's to the site

jburke-cadc:


jburke-cadc:
This pull request depends two other pull requests:
https://github.com/opencadc/core/pull/164
https://github.com/opencadc/reg/pull/109

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -101,11 +114,14 @@
...
 
-    // set init initConfig, used by subsequent init methods
+    static final int AVAILABILITY_CHECK_TIMEOUT = 30; //secs
+    static final int AVAILABILITY_FULL_CHECK_TIMEOUT = 300; //secs
 
+    // set init initConfig, used by subsequent init methods
     MultiValuedProperties props;
     File pubKey;
     File privKey;
+    private Thread availlabilityCheck;
```
pdowler, line 124: spelling: double `l`

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -173,6 +190,36 @@ private void initKeys() {
         log.info("initKeys: OK");
     }
 
+    private void initAvailabilityCheck() {
+        ArtifactDAO artifactDAO = new ArtifactDAO();
```
pdowler, line 194: Why create an ArtifactDAO here? The StorageSiteDAO copy ctor just copies the config map so you could instantiate it directly.

In principle, raven could run long enough that a new StorageSite appeared in the database. So the `list()` should get refreshed occasionally. So the list query should happen down in AvailabilityCheck.run(), right before the loop over known sites... doesn't need to happen every time, maybe same cadence of the full avail checks (5 min)

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -235,4 +282,91 @@ static MultiValuedProperties getConfig() {
             throw new IllegalStateException("invalid config: failed to load SQLGenerator: " + cname);
         }
     }
+
+    private class AvailabilityCheck implements Runnable {
+
+        private Set<StorageSite> sites;
+        private Map<URI, SiteState> siteStaties;
```
pdowler, line 289: spelling: siteStaties?? :-)

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -235,4 +282,91 @@ static MultiValuedProperties getConfig() {
...
+                        SiteState siteState = this.siteStaties.get(resourceID);
+                        if (siteState == null) {
+                            siteState = new SiteState(true, 0);
+                        }
+                        boolean minDetail = siteState.isMinDetail();
+                        Availability availability = getAvailability(resourceID, minDetail);
+                        siteState.available = availability.isAvailable();
+                        this.siteStaties.put(resourceID, siteState);
+                        this.siteAvailabilities.put(resourceID, availability);
+                        Context initContext = new InitialContext();
```
pdowler, line 320: I don't think you need to rebind the Map<URI,Availability> every time. It is a normal java object and you should be able to modify it without further interaction with JNDI (since you maintain a local reference).

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -235,4 +282,91 @@ static MultiValuedProperties getConfig() {
...
+                        initContext.rebind(JNDI_AVAILABILITY_KEY, this.siteAvailabilities);
+                        log.info(String.format("availability check [%s] for %s [%s]",
```
pdowler, line 322: I would log successful checks at debug and failed checks at warn. It would be nice to log "return to success" at info but it's probably more pain than it's worth.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -235,4 +282,91 @@ static MultiValuedProperties getConfig() {
...
+        private Set<StorageSite> sites;
+        private Map<URI, SiteState> siteStaties;
+        private Map<URI, Availability> siteAvailabilities;
+        
+        public AvailabilityCheck(Set<StorageSite> sites) {
+            this.sites = sites;
+            this.siteStaties = new HashMap<>(sites.size());
+            this.siteAvailabilities = new HashMap<>(sites.size());
+            try {
+                Context initialContext = new InitialContext();
```
pdowler, line 297: Here you need to be careful to check and unbind a previously bound object. yeah, it can't actually happen in the normal way of running raven in a container and always killing that and starting as new one, but the java code should be usable in a normal tomcat container, which means reloading the app without a tomcat restart.

In would also be good to be able to unbind when then thread gets terminated to not leave junk around.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -91,6 +103,7 @@
     private static final String RAVEN_KEY = "org.opencadc.raven";
 
     static final String JNDI_DATASOURCE = "jdbc/inventory"; // context.xml
+    static final String JNDI_AVAILABILITY_KEY = RAVEN_KEY + ".availabilities";
```
pdowler, line 106: RAVEN_KEY is for config items and not suitable for a JNDI key. The cadc-rest lib supplies a string that can be used as a prefix for JNDI keys (componentID).

pdowler (changes requested):

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
jburke-cadc: Fixed.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
jburke-cadc: Directly instantiated the StorageSiteDAO constructor, and moved the StorageSite query into the AvailabilityCheck run() method, where the query runs every AVAILABILITY_FULL_CHECK_TIMEOUT secs (currently 300).

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
jburke-cadc: Was having a bad spelling day apparently, fixed.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
jburke-cadc: Removed rebinding the availability Map every time it's updated.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
jburke-cadc: Updated the logging to debug at success and warn for failures.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
jburke-cadc: Updated to check for and remove an existing bind before binding the current object, and unbind on terminate.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
jburke-cadc: Used the appName from the InitAction & RestAction to generate the JNDI key.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
pdowler: Why this kind of check?  Why not just lookup and unbind? Why not just blindly unbind like up in the terminate method?

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -173,6 +192,38 @@ private void initKeys() {
...
+        StorageSiteDAO storageSiteDAO = new StorageSiteDAO();
+        storageSiteDAO.setConfig(getDaoConfig(props));
+
+        this.siteAvailabilitiesKey = this.appName + RavenInitAction.JNDI_AVAILABILITY_NAME;
+        terminate();
+        this.availabilityCheck = new Thread(new AvailabilityCheck(storageSiteDAO, this.siteAvailabilitiesKey));
+        this.availabilityCheck.setDaemon(true);
+        this.availabilityCheck.start();
+    }
+
```
pdowler, line 205: How does terminate() get called? 

I don't see a RavenInitAction.doShutdown() implementation anywhere...

pdowler (changes requested):

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
jburke-cadc: Yeah if you're going to unbind anyways, might as well just do the unbind. Updated.

Oops, missed the doShutdown(), added.

pdowler (approved):

PR #348, opened 2021-08-18
CADC-9716 - update the logging path to the servlet-name/artifact

jburke-cadc:
removing the auth token if present

### DIFF
--- a/minoc/build.gradle
+++ b/minoc/build.gradle
@@ -41,7 +41,8 @@ dependencies {
     compile 'org.opencadc:cadc-permissions-client:[0.2,)'
     compile 'org.opencadc:cadc-gms:[1.0,)'
     
-    compile 'org.opencadc:nom-tam-fits:[1.16.2,2.0)'
+    //compile 'org.opencadc:nom-tam-fits:[1.16.2,2.0)'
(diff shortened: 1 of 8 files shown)


pdowler (changes requested):
The changes here are tightly coupled to the order that RestServlet calls set methods. That's likely to be fragile. To fix that you need to override both setSyncInput() and setLogInfo():

* In both, you call the super impl first and never try to reproduce what it does in local code.

* In setSyncInput, you would call parsePath() and collect the correct `loggablePath` string and artifact uri; if logInfo is not null, call setPath.

* In setLogInfo, if `loggablePath` is not null, call setPath.

That way the order of those set calls in RestServlet doesn't matter.

Additionally, since these overridden methods come before the logging of  `longInfo.start()`, they cannot throw exceptions. For example, the IllegalArgumentException in parsePath cannot be thrown there as it will not result in the proper error handling. I would keep the initAndAuthorize() the way it was as it was and put the check for `if (artifactURI == null)` in the old init() method now that parsePath() happens before the action is invoked.

jburke-cadc:
That makes much more sense. Updated to follow the above pattern, and to not throw any exceptions before logInfo.start().

pdowler (approved):

PR #353, opened 2021-09-08
CADC-10140 add num_copies() function to luskan

jburke-cadc:


luskan/src/main/java/org/opencadc/luskan/NumCopiesConverter.java
```diff
@@ -0,0 +1,119 @@
...
+import net.sf.jsqlparser.expression.Function;
+import net.sf.jsqlparser.expression.operators.relational.ExpressionList;
+import net.sf.jsqlparser.schema.Column;
+import net.sf.jsqlparser.schema.Table;
+import org.apache.log4j.Logger;
+
+/**
+ * Class to change a num_copies() function to cardinality(inventory.artifact.sitelocations).
+ */
+public class NumCopiesConverter extends ExpressionNavigator {
```
pdowler, line 85: I would name this class InventoryFunctionConverter because it is highly likely we'll add another function to mangle in `visit` at some point.

And in visit: you should use mixed case table and column names got consistency with the naming style in the db. PG isn't case sensitive but other db servers are (in various ways). See comment in test code below.

luskan/src/main/java/org/opencadc/luskan/NumCopiesNavigator.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public void visit(PlainSelect ps) {
+        log.debug("visit(PlainSelect) " + ps);
+
+        List<Table> tables = new ArrayList<>();
+        List<Table> fromTableList = ParserUtil.getFromTableList(ps);
+        for (Table fromTable : fromTableList) {
+            if (fromTable.getName().equalsIgnoreCase("artifact")) {
```
pdowler, line 106: There could be a table named `tap_upload.artifact` in the query, so you have to look for `inventory.artifact` (that's the name in the tap_schema, so that's the name queries have to use).

luskan/src/test/java/org/opencadc/luskan/NumCopiesConverterTest.java
```diff
@@ -0,0 +1,170 @@
...
+        Log4jInit.setLevel("org.opencadc.luskan", Level.INFO);
+        Log4jInit.setLevel("net.sf.jsqlparser", Level.INFO);
+    }
+
+    private static final TapSchema tapSchema = TestUtil.mockTapSchema();
+
+    @Test
+    public void testSelect() {
+        String query = "select *, num_copies() from inventory.Artifact where id = '{some artifact id}'";
+        String expected = "SELECT inventory.Artifact.id, inventory.Artifact.contentLength, cardinality(inventory.artifact.sitelocations) FROM inventory.Artifact WHERE id = '{some artifact id}'";
```
pdowler, line 99: The AllColumnsConverter (for *) clearly preserves the case of the table name from the tap_schema, so you should expect `num_copies(inventory.Artifact.siteLocations)` here... and fix code to make that work. 

Well, there won't be a fix needed because the test in `doTest` does a toLowercase (which is common in this scenario - it gets you past a lot of needless discrepancies)... I'd probably add an additional arg to doTest like 
`String expectedCardinalityFunc` 
and do an 
`Assert.assertTrue(sql.contains(expectedCardinalityFunc))` 
before the toLowerCase

pdowler (changes requested):

luskan/src/main/java/org/opencadc/luskan/NumCopiesConverter.java
jburke-cadc: Changed the class name, and updated the schema.table.column names to match the db naming style.

luskan/src/main/java/org/opencadc/luskan/NumCopiesNavigator.java
jburke-cadc: The loop is looking for all Artifact tables. If there are multiple Artifact tables the methods exits without doing any work. But there does need to be a check if there's a single table in the FROM, it is an inventory.Artifact table.

luskan/src/test/java/org/opencadc/luskan/NumCopiesConverterTest.java
jburke-cadc: Updated the tests to check for the camel case argument to cardinality before doing the ignore case check of the SQL.

luskan/src/main/java/org/opencadc/luskan/NumCopiesNavigator.java
```diff
@@ -0,0 +1,144 @@
...
+ * Checks there is a single Artifact table in the FROM clause,
+ * then applies the InventoryFunctionConverter to the SELECT, WHERE, and GROUP BY clauses.
+ *
+ */
+public class NumCopiesNavigator extends SelectNavigator {
+    private static final Logger log = Logger.getLogger(NumCopiesNavigator.class);
+
+    InventoryFunctionConverter inventoryFunctionConverter;
+
+    public NumCopiesNavigator(InventoryFunctionConverter en, ReferenceNavigator rn, FromItemNavigator fn) {
```
pdowler, line 94: Because they are tightly coupled, NumCopiesNavigator should only have a no-arg ctor and create it's sub-components itself.

pdowler (changes requested):

luskan/src/main/java/org/opencadc/luskan/NumCopiesNavigator.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public void visit(PlainSelect ps) {
+        log.debug("visit(PlainSelect) " + ps);
+
+        List<Table> tables = new ArrayList<>();
+        List<Table> fromTableList = ParserUtil.getFromTableList(ps);
+        for (Table fromTable : fromTableList) {
+            if (fromTable.getName().equalsIgnoreCase("Artifact")) {
```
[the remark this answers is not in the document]
jburke-cadc: Okay so all Artifact tables are not necessarily the same. Updated the comparison to use the schema.table name.

luskan/src/main/java/org/opencadc/luskan/NumCopiesNavigator.java
```diff
@@ -0,0 +1,144 @@
...
+                tables.add(fromTable);
+                log.debug("found fromTable: " + fromTable.getWholeTableName());
+            }
+        }
+
+        // if no artifact tables found, return.
```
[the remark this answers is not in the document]
jburke-cadc: Updated, and renamed, the navigator to simply pass the list of tables in the from to the converter, then apply the converter to the SELECT, WHERE, and GROUP BY. The logic is in the converter, to check if the function can be applied to the query, and throw an exception if it can not.

luskan/src/test/java/org/opencadc/luskan/InventoryFunctionConverterTest.java
```diff
@@ -0,0 +1,180 @@
...
+        doTest(query, expected, null);
+    }
+
+    @Test
+    public void testNotInventorySchema() {
+        String query = "select count(*), num_copies() from temp.Artifact";
+        String expected = "select count(*), num_copies() from temp.Artifact";
+        doTest(query, expected, null);
+    }
+
```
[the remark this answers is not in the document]
jburke-cadc: Added the tests.

luskan/src/main/java/org/opencadc/luskan/NumCopiesNavigator.java
jburke-cadc: Updated the navigator constructor to only take the converter.

luskan/src/test/java/org/opencadc/luskan/InventoryFunctionConverterTest.java
```diff
@@ -0,0 +1,220 @@
...
+        String query = "select count(*), num_copies() from inventory.Artifact as a";
+        String expected = "select count(*), cardinality(a.siteLocations) from inventory.Artifact as a";
+        doTest(query, expected, "a.siteLocations");
+    }
+
+    @Test
+    public void testMultipleArtifactTables() {
+        String query = "select count(*), num_copies() from inventory.Artifact as a, temp.Artifact as b where a.id = b.id";
+        String expected = "select count(*), cardinality(a.siteLocations) from inventory.Artifact as a, temp.Artifact as b where a.id = b.id";
+        doTest(query, expected, null);
```
pdowler, line 128: could have `"a.siteLocations"` as expected cardinality here... I was expecting 
`"cardinality(a.siteLocations)"` as the string to look for in all these cases, just to be as explicit as possible.

luskan/src/test/java/org/opencadc/luskan/InventoryFunctionConverterTest.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Test
+    public void testNoInventoryArtifactTable() {
+        String query = "select num_copies() from DeletedArtifactEvent";
+        doTestFailure(query);
+    }
+
+    @Test
+    public void test() {
```
pdowler, line 156: Is this supposed to fail because it is `from DeletedArtifactEvent` or because `num_copies()` is in the group by? The method name doesn't express intent and it could fail for 3 different reasons and thus not be testing what it's supposed to (3rd: the mock tap_schema probably includes inventory.DeletedArtifactEvent).

Observe this:
```
cadctest=> select count(*),cardinality(siteLocations) from inventory.Artifact group by cardinality(siteLocations);
 count | cardinality 
-------+-------------
     3 |            
```
which says there are 3 artifacts with null siteLocations... so function can appear in group by without an alias and should be converted there. This test appears to be a duplicate "wrong table" test.

So, please add a `testGroupBy` (without alias) and verify that both uses of num_copies() get converted. This `test()` doesn't seem to test anything new.

luskan/src/main/java/org/opencadc/luskan/AdqlQueryImpl.java
```diff
@@ -105,7 +105,10 @@ protected void init() {
 
         super.navigatorList.add(new FunctionExpressionExtractor(
                 new PgFunctionNameConverter(), new ReferenceNavigator(), new FromItemNavigator()));
-        
+
+        // convert num_copies() -> cardinality(inventory.artifact.sitelocations)
+        super.navigatorList.add(new InventoryFunctionNavigator(new InventoryFunctionConverter()));
```
pdowler, line 110: I think you could move the `new InventoryFunctionConverter()` into the InventoryFunctionNavigator ctor (right in the super(...) like the others.

pdowler (changes requested):

luskan/src/test/java/org/opencadc/luskan/InventoryFunctionConverterTest.java
jburke-cadc: Missed that, the expected function should be been given, fixed.

luskan/src/test/java/org/opencadc/luskan/InventoryFunctionConverterTest.java
jburke-cadc: The jsqlparser will not parse a query with a function name in the group by.
`select count(*), num_copies() from inventory.Artifact group by num_copies()`
will throw a JSQLParserException on `()` in `group by num_copies()` 

The test wasn't named because it was a throw away to confirm a function can not be used in group by, and should have been deleted before the commit.

luskan/src/main/java/org/opencadc/luskan/AdqlQueryImpl.java
jburke-cadc: Moved the creating the InventoryFunctionConverter into the InventoryFunctionNavigator constructor.

pdowler (approved):

PR #355, opened 2021-09-17
CADC-10140 CADC-10135 if ratik finds a nonpolicy Artifact in local storage

jburke-cadc:
it will confirm multiple copies in global before deleting

### DIFF
--- a/ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
+++ b/ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
@@ -164,19 +164,19 @@ public class ArtifactValidator {
     protected void validateLocal(Artifact local)
         throws InterruptedException, IOException, ResourceNotFoundException, TransientException {
 
-        // explanation0: filter policy at L changed to exclude artifact in R
+        // explanation0: filter policy at L excludes artifact in R
         // evidence: R uses a filter policy AND Artifact in R without filter
         // if (L==global) delete Artifact, if (L==storage) delete Artifact only if
         // remote has multiple copies and create DeletedStorageLocationEvent
         log.debug("checking explanation 0");
         if (this.artifactSelector.getConstraint() != null) {
-            Artifact remote = getRemoteArtifact(local.getURI());
-            if (remote != null) {
+            ArtifactQueryResult queryResult = getRemoteArtifactQueryResult(local.getID());
+            if (queryResult != null && queryResult.artifact != null) {
                 boolean multipleCopies = false;
                 // if L == storage, get the Artifact count from global
                 if (this.remoteSite == null) {
-                    Integer count = this.getRemoteArtifactCount(local.getID());
-                    if (count != null && count > 1) {
+                    Integer numCopies = queryResult.numCopies;
+                    if (numCopies != null && numCopies > 1) {
                         multipleCopies = true;
                     }
                 }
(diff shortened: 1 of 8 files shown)


ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -626,6 +651,32 @@ Artifact getRemoteArtifact(URI uri)
         return null;
     }
 
+    Integer getRemoteArtifactCount(URI uri)
```
pdowler, line 654: This has to be a query by Artifact.id -- if the local artifact was recently PUT here it could have a different ID than the old (overwritten) entity in global... the num_copies() of the local artifact could still be 0/null/not-found (fenwick not yet propagated) while the old one might have 2+ copies. That would trigger a delete of the wrong (newer) artifact. 

context: in the design, a site with an artifact-filter that excludes an artifact can still end up accepting a put of such a file (eg acting as a cache if writes to other "preferred" sites fail)... that's how one could get an "unwanted" local artifact but have to wait until "that artifact" (not an artifact with the same uri) is propagated and the delete is safe.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
```diff
@@ -160,26 +162,49 @@ protected void validateLocal(Artifact local)
...
-                    log.info(String.format("ArtifactValidator.createDeletedStorageLocationEvent event=%s reason=local-filter-policy-change",
-                                           deletedStorageLocationEvent));
+                    DeletedStorageLocationEvent deletedStorageLocationEvent = new DeletedStorageLocationEvent(local.getID());
+                    log.info(String.format(
+                        "ArtifactValidator.createDeletedStorageLocationEvent event=%s reason=local-filter-policy-change",
+                        deletedStorageLocationEvent));
                     this.deletedStorageLocationEventDAO.put(deletedStorageLocationEvent);
+                } else {
+                    log.info(String.format(
+                        "ArtifactValidator.didNotDeleteArtifact Artifact.id=%s Artifact.uri=%s "
```
pdowler, line 205: if you look in tantar (InventoryIsAlwaysRight), there is a "no-op" event type there that is comparabel to the situation here. ".delayAction" that seems to be roughly the same meaning as here: it's too soon to do this action, so leaving this discrepancy alone for now.

the value of each k=v in the log message needs to be a single string (no punctuation or spaces) in order to be searchable. Here, you just need `reason=local-filter-policy-change` without the "but".

pdowler (changes requested):

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
jburke-cadc: Okay I see the difference, updated to query using the ID.

ratik/src/main/java/org/opencadc/ratik/ArtifactValidator.java
jburke-cadc: Updated the log message.

jburke-cadc: The builds for cadc-storage-adapter-ad and critwall were failing on the new RangeNotSatisfiableException in the cadc-util net classes. Included the fixes with this PR.

pdowler (approved):

PR #359, opened 2021-09-24
Cadc 8681 - GET with ranges support

andamian:


minoc/src/main/java/org/opencadc/minoc/GetAction.java
```diff
@@ -247,4 +282,51 @@ boolean hasNoOperations() {
             return conflicts;
         }
     }
+
+    /**
+     * Range is outside the file content
+     */
```
pdowler, line 288: Partial review only because this part will reshape the other code.

This exception must be defined in cadc-util (as RangeNotSatisfiableException to match the status code definition) *and* added to HttpTransfer.checkErrors to map a 416 to this exception. That can be reviewed and published.

Once that is done, it must be added to cadc-rest RestAction's main catch that maps exceptions to codes (416 here). Then this code simply throws that exception.

Also, I think you need to set some of the headers from `HeaderAction.setHeaders`. In my put-txn work I am adding a `HeaderAction.setTransactionHeaders` to supplement the standard headers... here it might be more subtle than that so you might have to figure out a clean way to not set all the standard headers if some should not be set, but it looks like all are applicable/unchanged.

pdowler (changes requested):

minoc/src/main/java/org/opencadc/minoc/GetAction.java
andamian: Didn't know about this design rule. A bit weird that it's not only `HttpGet` that uses that exception.
PR: https://github.com/opencadc/core/pull/169

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -352,7 +353,12 @@ private void doSync() {
         HttpPost post = new HttpPost(transferURL, content, true);
         post.setConnectionTimeout(6000); // ms
         post.setReadTimeout(60000);      // ms
-        post.prepare();
+        try {
+            post.prepare();
+        } catch (RangeNotSatisfiableException unexpected) {
```
pdowler, line 358: this local catch and wrap in RuntimeException is not the approp style for this class. I know it won't happen in practice, but say there is a poorly behaved file service that throws a 416... this will cause the while FileSyncJob to terminate rather than just use one of the retry strategies. So this should just be thrown here and dealt with in the approp catch up in doSync. 

In practice, doSync will just abort the job (`catch(Exception ex)` on line 202) and this wrap in RuntimeException will just hide the real cause.

You also need to add this exception to the approp catch in syncArtifact; there I think the catch on line 430 is the right choice: log at debug and remove the offending URL. As it is now, the RangeNotSatisfiableException will be rethrown  at line~446 and cause the job to terminate.

This probably seems pedantic since FileSyncJob doesn't curently do range requests (it is going to in the near future), but the main thing is to classify failures and be able to easily change the reaction as we gradually improve this code... and the wrap-in-RuntimeException will make it hard to deal with this exception correctly.

pdowler (changes requested):
Comments on the CI fixes in critwall here. I'll review the minoc changes tomorrow.

minoc/src/main/java/org/opencadc/minoc/GetAction.java
```diff
@@ -134,15 +143,41 @@ public void doAction() throws Exception {
...
+                        SortedSet<ByteRange> rangeSet = parseRange(range, artifact.getContentLength());
+                        if (rangeSet != null) {
+                            ByteRange byteRange = rangeSet.first();
+                            syncOutput.setCode(206);
+                            long lastByte = byteRange.getOffset() + byteRange.getLength() - 1;
+                            syncOutput.setHeader(CONTENT_RANGE, "bytes " + byteRange.getOffset() + "-"
+                                    + lastByte + "/" + artifact.getContentLength());
+                            // override content length
+                            syncOutput.setHeader(CONTENT_LENGTH, byteRange.getLength());
+                            storageAdapter.get(storageLocation, bcos, rangeSet);
```
pdowler, line 167: If the code gets in here, both this storageAdapter.get(...) and the one below on line 176 are going to be executed. I think this one is buried too deeply in conditionals and instead there should be a simpler block like this:
```
if (!rangeSet.isEmpty()) {
  storageAdapter.get(storageLocation, bcos, rangeSet);
} else {
  storageAdapter.get(storageLocation, bcos);
}
```
You could hide all the conditional complexity in parseRange and always return a (possibly empty) set.


The `catch(RangeNotSatisfiableException ex)` could be in the outer (line ~220) and set appropriate headers there (you don't need another layer of try/catch in here).

PR #360, opened 2021-09-28
CADC-9998 read optional put preferences in raven config

jburke-cadc:


raven/README.md
```diff
@@ -42,6 +42,14 @@ Multiple values of the permission granting service resourceID(s) may be provided
 settings. All services will be consulted but a single positive result is sufficient to grant permission for an 
 action.
 
+Another set of optional keys configure raven to prioritize sites returned in a transfer, with higher priority
+sites first in the list of transfer URL's to GET or PUT a file.
+```
+org.opencadc.raven.putPreference={storage site name}
+{storage site name).resourceID={storage site resourceID}
+{storage site name).namespaces={comma delimited list of namespaces}
```
pdowler, line 50: Don't use comma-delimited - the config supports multi-valued keys and you can get a List<String> without more parsing.

Also, `namespace` singular

raven/README.md
```diff
@@ -63,6 +71,10 @@ ivo://ivoa.net/std/UMS#login-0.1 = ivo://cadc.nrc.ca/gms
 
 ivo://ivoa.net/std/CDP#delegate-1.0 = ivo://cadc.nrc.ca/cred
 ivo://ivoa.net/std/CDP#proxy-1.0 = ivo://cadc.nrc.ca/cred
+
+org.opencadc.raven.putPreference=@CADC
+@CADC.resourceID=ivo://cadc.nrc.ca/cadc/minoc
+@CADC.namespaces=cadc:IRIS/ cadc:CGPS/
```
pdowler, line 77: This example doesn't even conform to the statement above, but will change further when multiple lines are used.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -286,6 +295,54 @@ static MultiValuedProperties getConfig() {
...
+    static Map<URI, StorageSiteRule> getStorageSiteRules(MultiValuedProperties props) {
+        StringBuilder sb = new StringBuilder();
+        Map<URI, StorageSiteRule> prefs = new HashMap<>();
+
+        List<String> putPreferences = props.getProperty("org.opencadc.raven.putPreference");
+        for (String putPreference : putPreferences) {
+
+            URI resourceID = null;
+            List<String> resourceIDs = props.getProperty(putPreference + ".resourceID");
+            if (resourceIDs.size() == 0) {
```
pdowler, line 307: isEmpty()

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -286,6 +295,54 @@ static MultiValuedProperties getConfig() {
...
+                sb.append(String.format("%s.resourceID: EMPTY VALUE\n", putPreference));
+            } else {
+                try {
+                    resourceID = new URI(resourceIDs.get(0));
+                } catch (URISyntaxException e) {
+                    sb.append(String.format("%s.resourceID: INVALID URI\n", putPreference));
+                }
+            }
+
+            String namespace = null;
```
pdowler, line 321: This bit below can be simplified by just using all values of {name}.namespace

raven/README.md
jburke-cadc: No it doesn't match the description, updated to use namespace with multiple lines.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -286,6 +295,51 @@ static MultiValuedProperties getConfig() {
...
+
+            List<String> namespaces = props.getProperty(putPreference + ".namespace");
+            if (namespaces.isEmpty()) {
+                sb.append(String.format("%s.namespace: MISSING\n", putPreference));
+            } else {
+                for (String namespace : namespaces) {
+                    if (!StringUtil.hasText(namespace)) {
+                        sb.append(String.format("%s.namespace: EMPTY VALUE\n", putPreference));
+                    } else if (namespace.matches(".*[\\s].*")) {
+                        sb.append(String.format("%s.namespace: MULTIPLE ENTRIES\n", putPreference));
```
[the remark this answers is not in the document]
jburke-cadc: Updated the error messages. An uppercase message previously was used since the message was a word or two long. That doesn't work well for a longer more informative message so I make them lower case. Not sure what value upper case brings to the logging. Easier to pick out maybe, but this is at startup when there isn't much logging to look through. 
We do check the syntax of the properties files, but we don't check the syntax of the property values. I added a Namespace class and a NamespaceSyntaxException class to raven for you to look at, permanent home TBD.
Baldur does not check the namespace syntax, other than making sure it's a valid regex expression. It does check that the groups are a valid GroupURI. I'll add an issue to baldur to check the namespace syntax.

raven/src/test/java/org/opencadc/raven/ProtocolsGeneratorTest.java
```diff
@@ -100,5 +107,51 @@ public void testPrioritizePullFromSites() throws Exception {
...
+        sites.add(new StorageSite(otherSite, "other", true, true));
+
+        Map<URI, StorageSiteRule> rules = new HashMap<>();
+        rules.put(cadcSite, new StorageSiteRule(cadcNS));
+        rules.put(readOnlySite, new StorageSiteRule(gemNS));
+        rules.put(otherSite, new StorageSiteRule(otherNS));
+
+        // artifact with no preferences in config, returns two read-write sites
+        List<StorageSite> ret = ProtocolsGenerator.prioritizePushToSites(sites, URI.create("get:SITE/file.ext"), rules);
+        Assert.assertEquals(2, ret.size());
```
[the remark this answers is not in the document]
jburke-cadc: Updated the test.

raven/src/test/java/org/opencadc/raven/ProtocolsGeneratorTest.java
```diff
@@ -100,5 +107,51 @@ public void testPrioritizePullFromSites() throws Exception {
...
+        Assert.assertEquals(cadcSite, ret.get(0).getResourceID());
+        Assert.assertEquals(otherSite, ret.get(1).getResourceID());
+
+        // artifact with namespace in read-only site, returns two read-write sites
+        ret = ProtocolsGenerator.prioritizePushToSites(sites, URI.create("gemini:GEMINI/file.ext"), rules);
+        Assert.assertEquals(2, ret.size());
```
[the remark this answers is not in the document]
jburke-cadc: Updated the test.

raven/src/main/java/org/opencadc/raven/StorageSiteRule.java
```diff
@@ -0,0 +1,86 @@
...
+ ************************************************************************
+ */
+
+package org.opencadc.raven;
+
+import java.util.List;
+
+public class StorageSiteRule {
+
+    private final List<String> namespaces;
```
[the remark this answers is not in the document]
jburke-cadc: Used the Namespace class for the list of namespaces.

raven/src/main/java/org/opencadc/raven/ProtocolsGenerator.java
```diff
@@ -244,14 +246,51 @@ static void prioritizePullFromSites(List<StorageSite> storageSites) {
...
+        List<StorageSite> orderedSites = new ArrayList<>();
+
+        // for each remote storage site in global
+        for (StorageSite storageSite : storageSites) {
+            // skip the storage site if it doesn't allow writes
+            if (!storageSite.getAllowWrite()) {
+                continue;
+            }
+
+            boolean added = false;
```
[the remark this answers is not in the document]
jburke-cadc: Added the comparator above. Should've thought of that in the first place. You just loop through the StorageSite's, ignore the readonly sites for the pushTo, add writable sites to the SortedSet, and then return the descending set to get the highest priority sites first.

raven/src/main/java/org/opencadc/raven/ProtocolsGenerator.java
```diff
@@ -341,4 +361,62 @@ private boolean isAvailable(URI resourceID) {
...
+            this.artifactURI = artifactURI;
+            this.clientIP = clientIP;
+        }
+
+        @Override
+        public int compare(StorageSite site1, StorageSite site2) {
+
+            boolean site1Match = hasMatchingNamespace(site1);
+            boolean site2match = hasMatchingNamespace(site2);
+
```
pdowler, line 386: First: the logic in the `hasMatchingNamespace` is wrong and that method is useless.

What you want here is something like:
```
// get namespace for each site
Namespace ns1 = siteRules.get(site1);
Namespace ns2 = siteRules.get(site2);

boolean site1Match = ns1.matches(artifactURI);
boolean site2Match = ns2.matches(artifactURI);
```
Above is pseudo-code and doesn't handle when ns1 or ns2 are null, which is possible when there are no rules for a site so it still needs some work. However, while reviewing Namespace it had occurred to me you would want a method like
```
public boolean matches(URI artifactURI)
```
and it is because I expected the above kind of usage.


Second: for the if/else if/else below, it is better to write it as:
```
if (site1Match && !site2match) {
     return 1;
}
if (!site1Match && site2match) {
     return -1;
}
// default: order of site names??
return site1.getResourceID().compateTo(site2.getResourceID());
```

raven/README.md
```diff
@@ -42,6 +42,14 @@ Multiple values of the permission granting service resourceID(s) may be provided
 settings. All services will be consulted but a single positive result is sufficient to grant permission for an 
 action.
 
+Another set of optional keys configure raven to prioritize sites returned in a transfer, with higher priority
+sites first in the list of transfer URL's to GET or PUT a file.
```
pdowler, line 46: The config options below only apply to PUT, but the above says GET or PUT.

The StorageSite class has a "name" attribute (it's just part of the resourceID and I'm kind of regretting it already) that might cause confusion with `{storage site name}` below. Maybe we should should use `{rule name}` (or `{entry name}` like in baldur)? 

Aside: over in baldur, we put the `{entry name} {regular expression}` on the same line, where we should have used 
```
{entry name}.pattern = {regular expression}
```
to re-use existing parsing functionality like here. That's one of those small style inconsistencies :-)

If we had done that, we would have noticed that the {pattern}'s in baldur and the {namespace}'s here are essentially **the same thing done two slightly different ways. :-(**

Also, the example below uses an @ sign which I know is just another character to be read as "at CADC", but I think it might be confusing and look like it's important when it isn't.

pdowler (changes requested):

raven/src/main/java/org/opencadc/raven/Namespace.java
```diff
@@ -0,0 +1,117 @@
...
+ *                                       <http://www.gnu.org/licenses/>.
+ *
+ *  : 5 $
+ *
+ ************************************************************************
+ */
+
+package org.opencadc.raven;
+
+public class Namespace {
```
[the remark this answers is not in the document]
jburke-cadc: Added javadoc for the class.

raven/src/main/java/org/opencadc/raven/Namespace.java
```diff
@@ -0,0 +1,117 @@
...
+
+    private final String namespace;
+
+    public Namespace(String value) throws NamespaceSyntaxException {
+        this.namespace = value;
+        parse(value);
+    }
+
+    protected void parse(String value) throws NamespaceSyntaxException {
```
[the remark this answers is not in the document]
jburke-cadc: Made the parse method private.

raven/src/main/java/org/opencadc/raven/Namespace.java
```diff
@@ -0,0 +1,117 @@
...
+        if (value == null) {
+            throw new NamespaceSyntaxException("null namespace");
+        }
+        if (value.isEmpty()) {
+            throw new NamespaceSyntaxException("zero length namespace");
+        }
+
```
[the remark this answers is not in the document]
jburke-cadc: Switched to using IllegalArgumentException instead of NamespaceSyntaxException.

raven/src/main/java/org/opencadc/raven/NamespaceSyntaxException.java
```diff
@@ -0,0 +1,85 @@
...
+ *                                       <http://www.gnu.org/licenses/>.
+ *
+ *  : 5 $
+ *
+ ************************************************************************
+ */
+
+package org.opencadc.raven;
+
+public class NamespaceSyntaxException extends Exception {
```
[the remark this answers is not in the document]
jburke-cadc: Deleted the class.

raven/src/main/java/org/opencadc/raven/ProtocolsGenerator.java
jburke-cadc: Added the `matches` method to Namespace, and rewrote the PrioritizingStorageSiteComparator compare to use the `matches` method.

raven/README.md
jburke-cadc: I updated the README to use `entry name` to match baldur. And removed the `@` before `CADC` in the example usage because yes one might wonder if it has some hidden meaning.

raven/src/main/java/org/opencadc/raven/ProtocolsGenerator.java
pdowler: Comparator still does order StorageSite instances correctly. Read the documentation and pay attention to the concept of "consistent with equals":. Specifically, if there is no StorageSiteRule for a two sites, that does not make them equal (line 401). It has to fall through to the default order. 

You clearly do not have unit tests that cover this scenario; if you did, one of those "equal" sites would not be added to the set and one URL would be missing from the resulting prioritised list of URLs.

raven/src/main/java/org/opencadc/raven/ProtocolsGenerator.java
```diff
@@ -244,14 +250,28 @@ static void prioritizePullFromSites(List<StorageSite> storageSites) {
...
+                                                        Map<URI, StorageSiteRule> siteRules) {
+        PrioritizingStorageSiteComparator comparator = new PrioritizingStorageSiteComparator(siteRules, artifactURI, null);
+        TreeSet<StorageSite> orderedSet = new TreeSet<>(comparator);
+        for (StorageSite storageSite : storageSites) {
+            if (storageSite.getAllowWrite()) {
+                orderedSet.add(storageSite);
+            }
+        }
+        // return the set in descending order, higher priority sites first.
+        return orderedSet.descendingSet();
```
pdowler, line 263: Why do you have to do this? Does the comparator order them backwards? Ugh.

pdowler (changes requested):
One issue replying to previous comment and one new one.

raven/src/main/java/org/opencadc/raven/ProtocolsGenerator.java
jburke-cadc: Missed that scenario. Updated the comparator to use the default StorageSite ordering when it is unable to determine the ordering between sites. I added a separate unit test for the the comparator class to make sure it is doing the right thing.

raven/src/main/java/org/opencadc/raven/ProtocolsGenerator.java
jburke-cadc: The comparator usually sorts ascending, higher ordered last in the set. The StorageSite ordering is also used, which would be ascending, so thought best to be consistent. But if the StorageSite ordering is being used the order isn't important. Changed the comparator to order descending for all but the default StorageSite ordering.

raven/src/test/java/org/opencadc/raven/ProtocolsGeneratorTest.java
```diff
@@ -100,5 +107,141 @@ public void testPrioritizePullFromSites() throws Exception {
...
+
+        //2nd StorageSite null
+        actual = comparator.compare(new StorageSite(URI.create("ivo:site1"), "site1", true, true), null);
+        Assert.assertEquals(-1, actual);
+
+        // 1st StorageSite null
+        actual = comparator.compare(null, new StorageSite(URI.create("ivo:site1"), "site1", true, true));
+        Assert.assertEquals(1, actual);
+
+
```
pdowler, line 132: Here you should have a compare test for the same two sites (2 and 5) and no rules so that you know it falls through to the StorageSite natural order and confirm the expected natural order. Otherwise, your test with a rule to "swap" the order below aren't necessarily working as intended.

In theory StorageSite could evolve and natural order could change, but since it is currently defined to use the resourceID that's a globally unique ID it seems unlikely. But: I still like to have a test that verifies the test conditions (it's not really an assumption). I wouldn't go so far as to make the test resilient to natural order changes because it's so unlikely and would obfuscate the test.

raven/src/test/java/org/opencadc/raven/RavenInitActionTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+package org.opencadc.raven;
+
+import ca.nrc.cadc.util.Log4jInit;
+import ca.nrc.cadc.util.PropertiesReader;
+import org.apache.log4j.Level;
+import org.apache.log4j.Logger;
+import org.junit.Assert;
+import org.junit.Test;
+
```
pdowler, line 78: This looks fine but I don't see why you need files for the keys if you don't call initKeys().

I never want to see keys/certs/etc committed because it always raises a warning flag. In future, I might actually read and verify the key files in initKeys() -- then they would have to be valid keys and that would be worse or might not be feasible. 

So, if the files are needed for the test to pass, then initKeys() is being called and that should be avoided.

pdowler (changes requested):

raven/src/test/java/org/opencadc/raven/RavenInitActionTest.java
jburke-cadc: The keys aren't needed for the tests. I think they originally were, before doInit() was overridden to call only the init methods being tested, and forgot to go back and cleanup the resources. Removed.

raven/src/test/java/org/opencadc/raven/ProtocolsGeneratorTest.java
```diff
@@ -100,5 +107,141 @@ public void testPrioritizePullFromSites() throws Exception {
...
+        ProtocolsGenerator.PrioritizingStorageSiteComparator comparator =
+            new ProtocolsGenerator.PrioritizingStorageSiteComparator(siteRules, URI.create("ivo:aaa"), clientIP);
+
+        // both StorageSite's null
+        int actual = comparator.compare(null, null);
+        Assert.assertEquals(0, actual);
+
+        //2nd StorageSite null
+        actual = comparator.compare(new StorageSite(URI.create("ivo:site1"), "site1", true, true), null);
+        Assert.assertEquals(-1, actual);
```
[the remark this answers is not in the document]
jburke-cadc: Updated tests to assert the result is not zero when the test should be use the default ordering .

raven/src/test/java/org/opencadc/raven/ProtocolsGeneratorTest.java
jburke-cadc: Added another test to with no site rules to confirm the default StorageSite order is used.

Not understanding the second paragraph, mainly `a test that verifies the test conditions`?

pdowler: In the case where you expect the rule to change the order, you need  the non-rule test first to make sure that isn't already the order; for example, if you had added the preference to site2 it would not change the order and your test wouldn't prove that rules change the order in the right way. 

So: one compare/assert for site2 vs site5 that verifies that site2 < site5 by default and then the real test w/ rule that makes site5 < site2. That lessens the chance that a bug goes undetected because one of those will fail if it isn't working as intended.

That means the asserts on line 147 and 152 *do care* whether the result is -1 or 1; all the others are OK.

pdowler (changes requested):

pdowler (approved):

PR #369, opened 2021-11-16
ratik out-of-policy cleanup improvement

pdowler:


ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
```diff
@@ -269,12 +276,21 @@ String buildRemoteQuery(final String bucket) throws ResourceNotFoundException, I
         DeletedStorageLocationEvent dsle = this.localEnvironment.deletedStorageLocationEventDAO.get(artifact.getID());
         Assert.assertNull("no remote: DeletedStorageLocationEvent not created", dsle);
 
+        // metaOnly should have been removed
```
jburke-cadc, line 279: Some of the code comments are not helping. 
Line 271 says `... if only a single copy in remote`.
There are no copies in remote, but still only a single copy in local, so the asserts are correct. Just initially confusing when the comment doesn't match the scenario.

Line 279 same issue, comment says metaOnly removed but assertion verifies it hasn't been removed?

jburke-cadc (approved):

ratik/src/intTest/java/org/opencadc/ratik/InventoryValidatorTest.java
pdowler: thnx. easy to stop looking at comments but they will be important next month when we come back in here :-)

PR #382, opened 2021-12-03
CADC-10552 split baldur properties .entry into .entry & {entry name}.pattern

jburke-cadc:


baldur/README.md
```diff
@@ -22,33 +22,38 @@ TODO: add support for allowedGroup to give service access.
 org.opencadc.baldur.allowedGroup = {groupURI}
 
 # one or more entry properties to grant permission to access artifacts
-# - each entry has a name and an Artifact URI pattern (regex)
+# - each entry has a name
+# - an Artifact URI pattern (regex)
```
pdowler, line 26: this comment still says "space separated list of" which is not correct: "zero or more group permissions" would be more accurate

pdowler (changes requested):

pdowler (changes requested):
baldur/VERSION needs to be updated to 0.2 to reflect the config change

pdowler (approved):

PR #389, opened 2022-01-17
update for API change in cadc-vos-1.2

pdowler:


minoc/src/main/java/org/opencadc/minoc/PutAction.java
```diff
@@ -273,6 +273,9 @@ public void doAction() throws Exception {
             profiler.checkpoint("transaction.commit.ok");
             log.debug("commit txn: OK");
             
+            syncOutput.setCode(201); // created
+            HeadAction.setHeaders(artifact, syncOutput);
+
```
brianmajor, line 278: Seeing HeadAction here threw me off a bit, but I see that it's the holder of the static method to set headers.  Maybe it could go in MinocAction instead but this is fine.

brianmajor (approved):
Looks good.

minoc/src/main/java/org/opencadc/minoc/PutAction.java
pdowler: yeah, probably not in the right place now that it's used more extensively

PR #397, opened 2022-02-03
CADC-10859 Removed archiveName and fileName from query. Trust uri fro…

yeunga:
…m the query. Updated unit tests.

### DIFF
--- a/cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
+++ b/cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
@@ -91,7 +91,7 @@ public class AdStorageQuery {
     private static final String QTMPL = "SELECT uri, inventoryURI, contentMD5, fileSize, ingestDate,"
             + " contentEncoding, contentType"
             + " FROM archive_files WHERE archiveName = '%s'"
-            + " ORDER BY ingestDate DESC";
+            + " ORDER BY uri ASC, ingestDate DESC";
 
     // some archive names are prefixes for others
     static final String DISAMBIGUATE_PREFIX = "x-";
--- a/cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
+++ b/cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
@@ -95,7 +95,7 @@ public class AdStorageQuery {
 
     // some archive names are prefixes for others
     static final String DISAMBIGUATE_PREFIX = "x-";
-    private static final List<String> ARC_PREFIX_ARC = Arrays.asList("CFHT", "GEM");
+    private static final List<String> ARC_PREFIX_ARC = Arrays.asList("CFHT", "GEM", "JCMT");
     
     private String query;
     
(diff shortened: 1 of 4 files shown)


cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
```diff
@@ -134,14 +134,17 @@ public AdStorageMetadataRowMapper() { }
...
-            String fname = (String) i.next();
             URI uri = (URI) i.next();
             if (uri == null) {
-                log.warn(AdStorageMetadataRowMapper.class.getSimpleName() + ".SKIP loc=" + archive + "/" + fname + " reason=null-uri");
+                log.warn(AdStorageMetadataRowMapper.class.getSimpleName() + ".SKIP loc=null/null reason=null-uri");
                 return null;
             }
             
+            // trust uri, <scheme>:<archive>/<fname>
+            String archiveFname = uri.toString().split(":")[1];
```
pdowler, line 144: The URI from line 137 is now the one that should be used directly as the storageID; it doesn't need to be deconstructed and reconstructed here. This applies to everything down to line ~153 where `URI storageID` is declared. 

The error message on line 139: the  "loc=null/null" is useless and should be removed; that warning message won't be very useful without a file-specific identifier but that's probably OK for now.

In the warning messages below, just use `"uri=" + uri` since that's never null.

pdowler (changes requested):

pdowler:
Also, can you add "JCMT" to the list of archive names that make ambiguous prefixes? (It's a constant list in AdStorageQuery)

cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
yeunga: Updated AdStorageQuery.

cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
```diff
@@ -134,24 +134,23 @@ public AdStorageMetadataRowMapper() { }
...
             URI uri = (URI) i.next();
             if (uri == null) {
-                log.warn(AdStorageMetadataRowMapper.class.getSimpleName() + ".SKIP loc=" + archive + "/" + fname + " reason=null-uri");
+                log.warn(AdStorageMetadataRowMapper.class.getSimpleName() + ".SKIP reason=null-uri");
                 return null;
             }
             
+            // trust uri, <scheme>:<archive>/<fname>
+            String archiveFname = uri.toString().split(":")[1];
+            final String archive = archiveFname.split("/")[0];
```
pdowler, line 145: How about just this?
```
String archive = uri.getSchemeSpecificPart().split("/")[0];
```

cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
```diff
@@ -134,24 +134,23 @@ public AdStorageMetadataRowMapper() { }
...
+            
             // chose best storageID
-            URI sid = URI.create("ad:" + archive + "/" + fname);
-            if ("mast".equals(uri.getScheme())) {
-                sid = uri;
-            }
+            URI sid = uri;
```
pdowler, line 148: This `sid` variable is now superfluous; in fact, you could now declare storageID way up at line 137. The easy way to do that is to remove lines 147-149, then rename `uri` to `storageID` on 137 (with IDE refactoring tool, it will rename all uses as well).

pdowler (changes requested):

cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
yeunga: Yes, sorry, overlooked.

cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
yeunga: Done.

PR #409, opened 2022-03-21
Cadc 11056 Add more log outputs to FileSyncJob.END

yeunga:
"downloadAuth" has not been tested yet. Will test it when the black network is available.

### DIFF
--- a/critwall/VERSION
+++ b/critwall/VERSION
@@ -1,4 +1,4 @@
 ## deployable containers have a semantic and build tag
 # semantic version tag: major.minor[.patch]
 # build version tag: timestamp
-TAGS="0.3-$(date --utc +"%Y%m%dT%H%M%S")"
+TAGS="0.4-$(date --utc +"%Y%m%dT%H%M%S")"
--- a/critwall/VERSION
+++ b/critwall/VERSION
@@ -1,4 +1,4 @@
 ## deployable containers have a semantic and build tag
 # semantic version tag: major.minor[.patch]
 # build version tag: timestamp
-TAGS="0.4-$(date --utc +"%Y%m%dT%H%M%S")"
+TAGS="0.3-$(date --utc +"%Y%m%dT%H%M%S")"
--- a/critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
+++ b/critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
@@ -138,8 +138,7 @@ public class FileSyncJob implements Runnable {
     private final Subject subject;
     private final Subject anonSubject = AuthenticationUtil.getAnonSubject();
     
-    private String downloadURL;
-    private String downloadAuth;
+    private String auth;
     private long byteTransferTime;
     private String artifactLabel;
     private final List<Exception> fails = new ArrayList<>();
@@ -168,8 +167,7 @@ public class FileSyncJob implements Runnable {
         
         this.artifactLabel = "Artifact.id=" + artifactID + " Artifact.uri=" + artifact.getURI();
         this.byteTransferTime = 0;
-        this.downloadURL = "";
-        this.downloadAuth = "";
+        this.auth = "";
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -345,11 +352,15 @@ private void doSync() {
...
         } finally {
             long dt = System.currentTimeMillis() - start;
+            long overheadTime = dt - byteTransferTime;
             StringBuilder sb = new StringBuilder();
             sb.append("FileSyncJob.END ").append(artifactLabel);
             sb.append(" success=").append(success);
-            sb.append(" duration=").append(dt);
             sb.append(" attempts=").append(syncArtifactAttempts);
+            sb.append(" downloadAuth=").append(downloadAuth);
+            sb.append(" downloadURL=").append(downloadURL);
```
pdowler, line 361: Critwall always uses the same auth during negotiation, so specifying `downloadAuth` specifically here is ugly for no pourpose. Just `auth`.

The `logURL` is a hack to strip off potential pre-auth tokens and log the target minoc service, so it's not consistent enough to add what I'd call "permanent logs". For success=true, it will be noise in the logs we don't care about and it won't be consistent enough over time to be searchable. For success=false, it will just be the last of several download URLs tried.

Please remove `downloaddURL` from here (just leave it in FileSyncJob.ERROR messages).

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -709,6 +721,7 @@ private String getLoggableString(URL url, URI sm) {
         }
         sb.append("...");
         if (sm != null) {
+            downloadAuth = sm.getFragment();
```
pdowler, line 724: Don't assign this as a side effect in this (hack) method. Assign independently around line~540.

Now that there is a separate `auth` log property, remove it from the logURL and log the property consistently in FileSyncJob.ERROR messages as well.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -138,6 +138,9 @@
     private final Subject subject;
     private final Subject anonSubject = AuthenticationUtil.getAnonSubject();
     
+    private String downloadURL;
```
pdowler, line 141: Should be able to remove `downloadURL` member var after other requested changes.

pdowler (changes requested):

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
```diff
@@ -345,11 +352,15 @@ private void doSync() {
...
             long dt = System.currentTimeMillis() - start;
+            long overheadTime = dt - byteTransferTime;
             StringBuilder sb = new StringBuilder();
             sb.append("FileSyncJob.END ").append(artifactLabel);
             sb.append(" success=").append(success);
-            sb.append(" duration=").append(dt);
             sb.append(" attempts=").append(syncArtifactAttempts);
+            sb.append(" downloadAuth=").append(downloadAuth);
+            sb.append(" downloadURL=").append(downloadURL);
+            sb.append(" byteTransferTime=").append(byteTransferTime);
```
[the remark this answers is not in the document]
yeunga: Since overheadTime works hand in hand with byteTransferTime, I'll do the same for overheadTime as well.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
yeunga: Replaced "downloadAuth" with "auth". Removed "downloadURL" from FileSyncJob.END.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
yeunga: Added "auth" to line ~540. Updated logURL and added "auth=" to FileSyncJob.ERROR messages.

critwall/src/main/java/org/opencadc/critwall/FileSyncJob.java
yeunga: Yes, removed.

critwall/VERSION
```diff
@@ -1,4 +1,4 @@
 ## deployable containers have a semantic and build tag
 # semantic version tag: major.minor[.patch]
 # build version tag: timestamp
-TAGS="0.3-$(date --utc +"%Y%m%dT%H%M%S")"
```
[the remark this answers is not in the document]
yeunga: Reverted VERSION change.

pdowler (changes requested):
In order to make byte transfer time meaningful, it needs to be reset to 0 at the start of an attempt. I would declare a local var outside the `for` loop and accumulate (line~596) in the local variable.

Then, assign the result to `byteTransferTime` right before the return statement (there are 2: line 600 and 612 depending on whether transaction is in use or not... that's a little ugh but the alternative was worse). Assigning right at `return` means all transfer failure modes are taken in to account (inc. verifyMetadata and commit).

Finally, up in the logging: please put "duration" back; that's the duration of the whole job and should still be logged. Then change the labels for the two new things to simply "transfer" and "overhead".

pdowler (approved):

PR #417, opened 2022-05-03
Added support for reducing the effects of eventual consistency (CADC-11250)

andamian:


pdowler (changes requested):
I see changes in config but no update of the README to describe them so this review will be minimal and just about config.

Raven currently has no default behaviours you can change with optional config and I prefer to keep it explicit if at all possible. It has optional config when no config means something sensible (eg you can have 0..n grant providers). 

I don't know why we used the term `notFoundDefault` in the story description because this is configuring behaviour and if you configure it then it isn't a default. Since what I've seen so far is already confusing me we're obviously not characterising this right. That happens a lot - something make sense (during story development) until you see some code and then it doesn't. Maybe if it was documented I could tell what you think it means... and/or we should discuss.

raven/src/main/java/org/opencadc/raven/ArtifactAction.java
```diff
@@ -185,6 +188,13 @@ protected ArtifactAction() {
       
         // get the storage site rules
         this.siteRules = RavenInitAction.getStorageSiteRules(props);
+        String cnf = props.getFirstPropertyValue(RavenInitAction.ARTIFACT_NOT_FOUND_DEFAULT_KEY);
```
pdowler, line 191: Just a hbeads up: there should never be case-conversion when reading config; that will introduce inconsistency and confusion. However, don't fix this yet as I haven't decided on the configuration concept that matches this feature yet but it will probably change.

pdowler (changes requested):

raven/src/intTest/java/org/opencadc/raven/FilesTest.java
```diff
@@ -296,16 +295,134 @@ public Object run() throws Exception {
         }
     }
 
-    public void checkHeadResult(HttpGet request, URI artifactURI, URI checksum, String contentType) {
+    @Test
+    public void testConsistencyHEAD() throws Exception {
```
pdowler, line 299: Since this method tests both HEAD and GET, it could be renamed, maybe testConsistencyPreventNotFound?

raven/src/intTest/java/org/opencadc/raven/FilesTest.java
```diff
@@ -110,6 +122,8 @@ public class FilesTest extends RavenTest {
     private static final URI RESOURCE_ID1 = URI.create("ivo://negotiation-test-site1");
     private static final URI RESOURCE_ID2 = URI.create("ivo://negotiation-test-site2");
 
+    private static final URI CONSIST_RESOURCE_ID = URI.create("ivo://cadc.nrc.ca/minoc");
```
pdowler, line 125: This is the default local minoc test instance: that's fine. However, it is critical that this service and the raven service being tested use a different backend database. Otherwise, when the PUT in the consistency test finished, the artifact will be in raven's database (presumably without a siteLocation)... not sure what would happen but it would be easy for incorrect test system setup to be really confusing. I sometimes re-use the back end database for different things and while I do have a separate minoc db it was for something else and this is a subtle new requirement in raven tests.

Please update src/intTest/README.md to make it clear that `ivo://cadc.nrc.ca/minoc` and `ivo://cadc.nrc.ca/raven` *must* have distinct back end database.

pdowler (changes requested):
Couple of consmetic things.

I think preventNotFound also needs to be tested in NegotiationTest... that calls ProtocolsGenerator with some more general user-supplied values (specifically, multiple protocols) that FilesAction doesn't really hit.

pdowler (approved):

PR #424, opened 2022-05-25
Added support for buckets in ad adapter CADC-1096

andamian:


pdowler (changes requested):
Also need:

1. additional test method for a vault file added to AdStorageAdapterGetTest to make sure vault files can be retrieved. Should use one of the existing static test files (a public one for simplicity).
2. it would be nice to have a sub-archive bucket test in AdStorageAdapterIteratorTest, but the 3-char limit would in production VOSpac makes the minimum query result about 62K rows... the logic should work with `IRIS:0` which would be 1/16 of IRIS so that might make a good test.

cadc-storage-adapter-ad/src/intTest/java/org/opencadc/inventory/storage/ad/AdStorageAdapterGetTest.java
```diff
@@ -149,6 +168,30 @@ public void testGetIgnoresStorageBucket() {
...
+        final URI expectedVOSpacChecksum = URI.create("md5:b05403212c66bdc8ccc597fedf6cd5fe");
+        Subject.doAs(testSubject, (PrivilegedExceptionAction<Object>) () -> {
+                final OutputStream outputStream = new ByteArrayOutputStream();
+                final DigestOutputStream digestOutputStream = new DigestOutputStream(outputStream, MessageDigest
+                        .getInstance(AdStorageAdapterGetTest.DIGEST_ALGORITHM));
+                final ByteCountOutputStream byteCountOutputStream = new ByteCountOutputStream(digestOutputStream);
+                final MessageDigest messageDigest = digestOutputStream.getMessageDigest();
+
+                final StorageLocation storageLocation = new StorageLocation(TEST_VOSPAC_URI);
+                storageLocation.storageBucket = AdStorageQuery.DISAMBIGUATE_PREFIX + "VOSpac:b054";
```
pdowler, line 185: I don't get why the test StorageLocation includes the DISAMBIGUATE. That `x-` is used in cases where one archive name is a prefix for another (eg GEM and GEMINI). It harmless since it gets removed by `bucket2archive` but it seems misleading here... makes one think it is necessary when it isn't.

cadc-storage-adapter-ad/src/intTest/java/org/opencadc/inventory/storage/ad/AdStorageAdapterIteratorTest.java
```diff
@@ -208,6 +204,78 @@ public Object run() throws Exception {
         });
     }
 
+    @Test
+    public void testIteratorVOSpac() throws Exception {
+        Subject.doAs(testSubject, new PrivilegedExceptionAction<Object>() {
+
+            // Should be VOSpac except that the buckets in that archive are too large
```
pdowler, line 211: this is quite complex to look at and the overridable method is ugh... now that we are really doing 4-char buckets, a single 4 char bucket has ~3700 files, which is about the same as the IRIS test in `testIteratorOrder()`... that would let you get rid of `getVOSpac()` and use a constant (or, to be honest, I don't mind using hard coded strings that are only in one class... sometimes easier to see exactly what's going on)

cadc-storage-adapter-ad/src/main/java/org/opencadc/inventory/storage/ad/AdStorageQuery.java
```diff
@@ -120,12 +146,21 @@ public TapRowMapper<StorageMetadata> getRowMapper() {
     //    }
     //    return arc;
     //}
-    
+
+    String getVOSpac() {
```
pdowler, line 150: see comment in the iterator test; I don't think you really need this method and can use `"VOSpac"` down on line ~160.

pdowler (changes requested):

cadc-storage-adapter-ad/src/intTest/java/org/opencadc/inventory/storage/ad/AdStorageAdapterIteratorTest.java
andamian: Makes sense. Tested it and there are no significant performance hits in the int test.

pdowler (approved):

PR #445, opened 2022-09-12
Support stream cutouts for large FZ (compressed) files

at88mph:
Add integration test with 310MB `test-cfht.fz` file.  Requires `opencadc:nom-tam-fits:1.16.6`.


pdowler (approved):

pdowler:
leaving PR unmerged for now (until I can run the full int-test suite) so I don't lose track

PR #452, opened 2022-10-17
tantar improvements

pdowler:
tries to move validation policy code to cadc-storage-adapter  to enable StorageAdapter.getPolicy() but it led to complex dependencies, so reverted to having all file-validation policy in tantar

cleaned up and renamed things in tantar policy code for clarity and long term ease of understanding

add tantar integration tests for 2 currently supported policies; did some more work (incomplete) on RecoverFromStoragePolicy


tantar/src/main/java/org/opencadc/tantar/Main.java
```diff
@@ -121,39 +127,81 @@ public static void main(final String[] args) {
...
+            ResolutionPolicy validationPolicy = null;
+            String policyClassName = props.getFirstPropertyValue(ResolutionPolicy.class.getName());
+            policyClassName = addPackage(policyClassName);
+            if (StringUtil.hasLength(policyClassName)) {
+                validationPolicy = InventoryUtil.loadPlugin(policyClassName);
+            } else {
+                throw new InvalidConfigException("required config property missing: " + ResolutionPolicy.class.getName());
+            }
+            
+            final String rawBucketRange = props.getFirstPropertyValue(BUCKETS_KEY);
```
jburke-cadc, line 154: Null or empty string check for the bucket range?

jburke-cadc (approved):

PR #463, opened 2022-11-25
add StorageAdapter.recover() method to API and implementations

pdowler:
add optional recovery test scenarios to test suite

### DIFF
--- a/cadc-storage-adapter-swift/src/main/java/org/opencadc/inventory/storage/swift/SwiftStorageAdapter.java
+++ b/cadc-storage-adapter-swift/src/main/java/org/opencadc/inventory/storage/swift/SwiftStorageAdapter.java
@@ -67,6 +67,7 @@
 
 package org.opencadc.inventory.storage.swift;
 
+import ca.nrc.cadc.date.DateUtil;
 import ca.nrc.cadc.io.ByteCountInputStream;
 import ca.nrc.cadc.io.ByteLimitExceededException;
 import ca.nrc.cadc.io.MultiBufferIO;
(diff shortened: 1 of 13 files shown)


cadc-storage-adapter-test/src/main/java/org/opencadc/inventory/storage/test/StorageAdapterBasicTest.java
```diff
@@ -445,4 +447,160 @@ public void testIteratorBucketPrefix() {
...
+            NewArtifact na = new NewArtifact(artifactURI);
+            na.contentLength = datalen;
+            StorageMetadata sm = adapter.put(na, TestUtil.getInputStreamOfRandomBytes(datalen), null);
+            Assert.assertFalse(sm.deletePreserved);
+            
+            
+            Thread.sleep(2000L);
+            adapter.delete(sm.getStorageLocation());
+            
+            try {
```
brianmajor, line 469: I don't see the point of the try/catch block... if the exception is expected, why is no action performed when the exception does not occur?

cadc-storage-adapter-swift/src/main/java/org/opencadc/inventory/storage/swift/SwiftStorageAdapter.java
```diff
@@ -1383,6 +1401,42 @@ public void delete(StorageLocation storageLocation, boolean includeRecoverable)
         }
         throw new ResourceNotFoundException("not found: " + storageLocation);
     }
+    
+    @Override
+    public void recover(StorageLocation storageLocation, Date contentLastModified) 
+            throws ResourceNotFoundException, IOException, InterruptedException, StorageEngageException, TransientException {
+        log.debug("delete: " + storageLocation);
```
brianmajor, line 1408: Check for null storageLocation here?

brianmajor (approved):

PR #484, opened 2023-01-09
CADC-11463 initial vault service with Dockerfile

jburke-cadc:


vault/src/main/java/org/opencadc/vault/VaultInitDatabase.java
```diff
@@ -0,0 +1,106 @@
...
+        "vospace.ModelVersion.sql",
+        "vospace.NodeAvailability.sql",
+        "vospace.Node.sql",
+        "vospace.NodeProperty.sql",
+        "vospace.DeletedNode.sql"
+    };
+
+    public static final String MODEL_NAME = "vospace";
+    public static final String MODEL_VERSION = "0.1";
+    public static final String PREV_MODEL_VERSION = "0.1";
```
pdowler, line 93: PREV version should be "n/a". Def not null (that means something "any" and is only useful in special cases).

vault/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+    }
+    from('.') {
+        include 'VERSION'
+    }
+}
+
+dependencies {
+    compile 'org.opencadc:cadc-log:[1.1.6,2.0)'
+    compile 'org.opencadc:cadc-vosi:[1.4.3,2.0)'
+    compile 'org.opencadc:cadc-inventory-db:[0.14,1.0)'
```
pdowler, line 31: should not include cadc-inventory-db dependency because it's not needed (yet)

must not include postgresql at runtime - in fact should be in the runtime.exclude

is all that slf4j stuff cruft copied from someplace else or is it actually being pulled in? If not actually needed, please remove.

vault/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## configuration
+See the [cadc-tomcat](https://github.com/opencadc/docker-base/tree/master/cadc-tomcat) image docs
+for expected deployment and general config requirements. The `vault` war file can be renamed
+at deployment time in order to support an alternate service name, including introducing 
+additional path elements (see war-rename.conf).
+
+Runtime configuration must be made available via the `/config` directory.
+
+### catalina.properties (cadc-tomcat)
```
pdowler, line 11: remove `(cadc-tomcat)` from this heading

remove documented SQLGenerator since it is not needed.

for connection pool, I would name props `org.opencadc.vault.nodes.{prop}` since a vospace will also need a `uws` pool and vospace is too vague. 

there is an unclosed triple-tickmark box below that is messing with formatting

pdowler (changes requested):
minor comments below

I didn't look very closely at the create table sql because it's pointless without the java code. I did note that the nodeID is a bigint. iirc, the code in the current db lib probably relies on that being a sequence in sybase and I don't recall a random bigint generator anywhere (since caom2-2.0 anyway) so I doubt those tables will actually work, but that's a problem for later :-)

vault/src/main/java/org/opencadc/vault/VaultInitDatabase.java
jburke-cadc: Updated to n/a.

vault/build.gradle
jburke-cadc: cadc-inventory-db is used to get the SQLGenerator class name, used in vault.properties. But the SQLGenerator property isn't currently used.
Better to leave in the dependency, or pull it and that property?
Copied over another SI build.gradle to start with, can't remember which one. slf4j isn't pulled in. Removed slf4j from the build.

pdowler: remove usage and from build

pdowler (changes requested):
I don't know how we got off on such a confusing track, but the Initdatabase and sql files do not belong in this module, even temporarily. They belong with the db library, which does not exist yet.

Please remove the VaultInitDatabase.java and all sql files; comment out the reference to VaultInitDatabase in VaultInitAction (which does belong and will do that once there is a db library to provide it).

Once again, I think this is "pressure to start working" wasting more time than it saves. Sorry for not noticing earlier.

jburke-cadc:
I think it was the rush to get a working container with database init for cadc-vos-server-db testing.
Removed the SQL and the database init class.

pdowler (approved):

PR #486, opened 2023-01-10
CADC-12121 move DBUtil from cadc-inventory-util to cadc-util

jburke-cadc:


critwall/build.gradle
```diff
@@ -23,7 +23,6 @@ dependencies {
     compile 'org.opencadc:cadc-util:[1.6,2.0)'
```
pdowler, line 23: probably should put the minimum cadc-util at 1.8.5 here

pdowler (changes requested):

pdowler (approved):

PR #524, opened 2023-07-25
Test updates and adapt to nom-tam-fits 1.18.0.

at88mph:
Depends on https://github.com/opencadc/dal/pull/95.

### DIFF
--- a/minoc/README.md
+++ b/minoc/README.md
@@ -148,16 +148,6 @@ docker run --rm -it minoc:latest /bin/bash
 docker run --rm --user tomcat:tomcat --volume=/path/to/external/config:/config:ro --name minoc minoc:latest
 ```
 
-## apply semantic version tags
-```bash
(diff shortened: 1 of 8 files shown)


minoc/README.md
```diff
@@ -157,3 +157,16 @@ done
 unset TAGS
 docker image list minoc
 ```
+
```
pdowler, line 160: The main README primarily serves as docs for the config so I'd like to avoid adding helpful things like this usage example... they also introduce a bunch of things that need explaining (cert support in front end proxy, permissions related config of minoc, potentially that identity belonging to a group with write permission to that namespace).

please remove; I will probably even remove the bit about tagging because it is not actually useful/interesting to developers or deployers (just publishers, and it isn't complete for them).

pdowler (changes requested):

pdowler (approved):

PR #543, opened 2023-12-11
Recursive properties (CADC-12562)

andamian:


vault/src/main/java/org/opencadc/vault/RecursiveNodePropsJobManager.java
```diff
@@ -0,0 +1,113 @@
...
+ * @author pdowler, majorb, yeunga, adriand
+ */
+public class RecursiveNodePropsJobManager extends SimpleJobManager {
+    private static final Logger log = Logger.getLogger(RecursiveNodePropsJobManager.class);
+
+    private static final Long MAX_EXEC_DURATION = Long.valueOf(12 * 7200L); // 24 hours?
+    private static final Long MAX_DESTRUCTION = Long.valueOf(7 * 24 * 3600L); // 1 week
+    private static final Long MAX_QUOTE = Long.valueOf(12 * 7200L); // same as exec
+
+    protected static JobPersistence jp;
```
pdowler, line 93: I don't know were this static JobPersistence style came from but it's a terrible idea and unnecessary. 

See cavern's `uws` package for one way to create them the same way (there are 4 in total)

pdowler (changes requested):

vault/src/main/java/org/opencadc/vault/RecursiveNodePropsJobManager.java
andamian: Fixed.

pdowler (approved):

PR #546, opened 2023-12-14
`raven` generates and persists pre-auth keys (CADC 12946)

andamian:


raven/src/intTest/java/org/opencadc/raven/NegotiationTest.java
```diff
@@ -835,4 +837,63 @@ public Object run() throws Exception {
             Assert.fail("unexpected exception: " + e);
         }
     }
+
+    @Test
+    public void testPreauthURL() throws Exception {
+        List<Protocol> requested = new ArrayList<>();
```
pdowler, line 843: probably don't need lines 843-846 and 875 below - that's for testing something else.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -150,6 +148,8 @@ void initDAO() {
         Map<String,Object> dc = getDaoConfig(props);
         ArtifactDAO artifactDAO = new ArtifactDAO();
         artifactDAO.setConfig(dc); // connectivity tested
+        PreauthKeyPairDAO kpDAO = new PreauthKeyPairDAO();
```
pdowler, line 151: this is not necessary since instantiating the artifactDAO is enough to check the db connection; that's all that's happening here

raven/src/main/webapp/WEB-INF/web.xml
```diff
@@ -106,6 +116,11 @@
     <servlet-name>FilesServlet</servlet-name>
     <url-pattern>/files/*</url-pattern>
   </servlet-mapping>
+
+  <servlet-mapping>
+    <servlet-name>PubKeyServlet</servlet-name>
+    <url-pattern>/pubkey/*</url-pattern>
```
pdowler, line 122: technically, do not need the trailing `/*` here, but probably harmless

raven/src/main/webapp/WEB-INF/web.xml
```diff
@@ -47,6 +47,16 @@
         <load-on-startup>2</load-on-startup>
     </servlet>
 
+  <servlet>
+    <servlet-name>PubKeyServlet</servlet-name>
+    <servlet-class>ca.nrc.cadc.rest.RestServlet</servlet-class>
+    <init-param>
+      <param-name>get</param-name>
+      <param-value>org.opencadc.raven.GetPubKeyAction</param-value>
```
pdowler, line 55: I don't see this class included in your PR... but I think we can re-use the one from vault after moving it to cadc-inventory-server. So don't fix right now... I'll try to integrate

pdowler (changes requested):
a few minor things... I will probably re-org/optimize the key pair setup to be more like vault after I integrate but nothing to do right now

raven/src/intTest/java/org/opencadc/raven/NegotiationTest.java
andamian: C&P error. Fixed

raven/src/main/webapp/WEB-INF/web.xml
andamian: Sorry - forgot to add the file to the commit but I have it locally if you need it.

PR #547, opened 2023-12-16
integrate work for vault transfer negotiation and magic preauth keys

pdowler:
requires unpublished external changes: cadc-registry cadc-vos-server cadc-test-vos so CI or testing might fail


minoc/src/main/java/org/opencadc/minoc/MinocConfig.java
```diff
@@ -0,0 +1,386 @@
...
+        return trustedServices;
+    }
+    
+    private void syncKeys() {
+        if (!trustedServiceKeySync) {
+            int numFails = 0;
+            RegistryClient reg = new RegistryClient();
+            // check map for null keys and try to retrieve them
+            for (Map.Entry<URI,byte[]> me : trustedServices.entrySet()) {
+                if (me.getValue() == null) {
```
andamian, line 315: I don't see how right now, the value is not `null` at this point. Can they be loaded from somewhere else?

minoc/src/main/java/org/opencadc/minoc/ServiceAvailability.java
```diff
@@ -213,6 +213,9 @@ public Availability getStatus() {
                 }
             }
 
+            // TODO: check grant providers
+            // TODO: check for null pubkeys for trusted services
+
```
andamian, line 218: Good point. You can check it here.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
```diff
@@ -184,18 +199,61 @@ void initStorageSiteRules() {
...
+    private void initKeyPair() {
+        log.info("initKeyPair: START");
+        jndiPreauthKeys = appName + "-" + PreauthKeyPair.class.getName();
+        try {
+            Map<String,Object> dc = getDaoConfig(props);
+            PreauthKeyPairDAO dao = new PreauthKeyPairDAO();
+            dao.setConfig(dc);
+            PreauthKeyPair keys = dao.get(KEY_PAIR_NAME);
+            if (keys == null) {
+                KeyPair kp = RsaSignatureGenerator.getKeyPair(4096);
```
andamian, line 211: So you are changing the DB account for `raven`?

andamian (approved):

minoc/src/main/java/org/opencadc/minoc/MinocConfig.java
```diff
@@ -0,0 +1,386 @@
...
+                }
+            }
+            trustedServiceKeySync = false;
+            syncKeys();
+        }
+        
+        List<String> recov = configProperties.getProperty(RECOVERABLE_NS_KEY);
+        if (recov != null) {
+            for (String s : recov) {
+            }
```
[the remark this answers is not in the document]
pdowler: incomplete code re-org. fixed

minoc/src/main/java/org/opencadc/minoc/MinocConfig.java
pdowler: the syncKeys() can fail and leave the map entry with null values (the ctor only adds keys with null values)... for example if minoc starts up before a trusted service or fails to get the pub key

I'm not that happy with the rigidity of the check, especially because it gets called on every request (with a preauth key in the URL)... bad config can effect every such call so it isn't great right now.

raven/src/main/java/org/opencadc/raven/RavenInitAction.java
pdowler: I decided to add a second pool for admin, which also allows raven to init the inventory tables instead of waiting for fenwick to have been run

PR #550, opened 2024-01-18
vault: add support for separate nodes and inventory pools

pdowler:
added inventory table init

### DIFF
--- a/vault/src/main/java/org/opencadc/vault/VaultInitAction.java
+++ b/vault/src/main/java/org/opencadc/vault/VaultInitAction.java
@@ -88,6 +88,7 @@ import org.opencadc.inventory.PreauthKeyPair;
 import org.opencadc.inventory.db.PreauthKeyPairDAO;
 import org.opencadc.inventory.db.SQLGenerator;
 import org.opencadc.inventory.db.StorageSiteDAO;
+import org.opencadc.inventory.db.version.InitDatabase;
 import org.opencadc.inventory.transfer.StorageSiteAvailabilityCheck;
 import org.opencadc.vospace.db.InitDatabaseVOS;
 import org.opencadc.vospace.server.NodePersistence;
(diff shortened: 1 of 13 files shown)


vault/src/main/java/org/opencadc/vault/VaultInitAction.java
```diff
@@ -227,14 +240,25 @@ static MultiValuedProperties getConfig() {
...
 
     static Map<String, Object> getDaoConfig(MultiValuedProperties props) {
-        
         Map<String, Object> ret = new TreeMap<>();
         ret.put(SQLGenerator.class.getName(), SQLGenerator.class); // not configurable right now
-        ret.put("jndiDataSourceName", org.opencadc.vault.VaultInitAction.JNDI_DATASOURCE);
+        ret.put("jndiDataSourceName", org.opencadc.vault.VaultInitAction.JNDI_VOS_DATASOURCE);
         ret.put("schema", props.getFirstPropertyValue(org.opencadc.vault.VaultInitAction.INVENTORY_SCHEMA_KEY));
         ret.put("vosSchema", props.getFirstPropertyValue(org.opencadc.vault.VaultInitAction.VOSPACE_SCHEMA_KEY));
         return ret;
```
jburke-cadc, line 248: Does the DaoConfig need both the `vospace` and `inventory` schema's? 
The schema key for both `DaoConfig` and `InvConfig` is the INVENTORY_SCHEMA_KEY.
Also shouldn't need the full class path to those keys, makes the line wrap and hard to decipher when doing a code review. :-)

jburke-cadc (changes requested):

vault/src/main/java/org/opencadc/vault/VaultInitAction.java
pdowler: ugh, I think my IDE did that when I pasted... fixing

PR #551, opened 2024-01-25
minoc: filename override mechanism

pdowler:


cadc-inventory-server/src/main/java/org/opencadc/inventory/transfer/ProtocolsGenerator.java
```diff
@@ -486,52 +508,50 @@ private List<Protocol> doPushTo(URI artifactURI, Transfer transfer, String authT
...
+                    }
+                    boolean anon = Standards.SECURITY_METHOD_ANON.equals(sec);
+                    Interface iface = filesCap.findInterface(sec);
+                    log.debug("pushTo: " + storageSite + " proto: " + proto + " iface: " + iface);
+                    if (iface != null) {
+                        URL baseURL = iface.getAccessURL().getURL();
+                        //log.debug("base url for site " + storageSite.getResourceID() + ": " + baseURL);
+                        if (protocolCompat(proto, baseURL)) {
+                            // // no plain anon URL for put: !anon or anon+token
+                            boolean gen = (!anon || (anon && authToken != null));
```
andamian, line 531: From what I'm seeing, the `pushTo` and `pullFrom` loops are pretty similar except this and the file name override. Would it be possible (or worth it) to avoid the code duplication?

cadc-inventory-server/src/test/java/org/opencadc/inventory/transfer/ProtocolsGeneratorTest.java
```diff
@@ -99,10 +100,16 @@ public void testPrioritizePullFromSites() throws Exception {
...
         }
-        ProtocolsGenerator.prioritizePullFromSites(sites);
-        for (StorageSite s : sites) {
-            log.info("found: " + s.getID() + " aka " +  s.getResourceID());
-        }
+        List<StorageSite> result1 = ProtocolsGenerator.prioritizePullFromSites(sites);
+        Assert.assertEquals(sites.size(), result1.size());
+        Assert.assertTrue(result1.containsAll(sites));
+        
+        List<StorageSite> result2 = ProtocolsGenerator.prioritizePullFromSites(sites);
```
andamian, line 107: What is this testing?

minoc/src/test/java/org/opencadc/minoc/ArtifactActionTest.java
```diff
@@ -147,10 +154,16 @@ public void testParsePath() {
...
-            assertCorrectPath("mast:long/uri/with/segments/fits.fits", "mast:long/uri/with/segments/fits.fits", null);
+            assertCorrectPath("mast:long/uri/with/segments/something.fits", "mast:long/uri/with/segments/something.fits", null);
             assertCorrectPath("token/mast:long/uri/with/segments/fits.fits", "mast:long/uri/with/segments/fits.fits", "token");
             assertCorrectPath("token-with-dashes/cadc:TEST/myartifact", "cadc:TEST/myartifact", "token-with-dashes");
             
+            assertCorrectPath("cadc:vault/uuid:fo/something.fits", "cadc:vault/uuid", null, "something.fits");
+            assertCorrectPath("token/cadc:vault/uuid:fo/something.fits", "cadc:vault/uuid", "token", "something.fits");
+            
+            assertCorrectPath("cadc:vault/uuid:/something.fits", "cadc:vault/uuid:/something.fits", null, null);
+
```
andamian, line 165: How does it handle the `cadc:vault/uuid:fo/` case?

andamian (approved):

cadc-inventory-server/src/test/java/org/opencadc/inventory/transfer/ProtocolsGeneratorTest.java
pdowler: it is just testing that pull from prioritization is random by those two sets not being the same order

minoc/src/test/java/org/opencadc/minoc/ArtifactActionTest.java
pdowler: this just tests that extra colons do not cause parts of the path to be lost (are preserved in the artifact uri); it is easy to implement the extraction in ways that lose stuff (eg if you split on colon and then it isn't a `:fo/`) 

the previous test checks that `:fo/` is extracted and becomes the override filename

andamian: Sorry for being too vague, what I've meant is the incorrect case in which the URL ends in `:fo/`.

cadc-inventory-server/src/main/java/org/opencadc/inventory/transfer/ProtocolsGenerator.java
pdowler: maybe, but iirc there were enough subtle differences that made that more complex. I think you originally wrote it and convinced me of that, although now there's plenty of `git blame` to go around :-)

cadc-inventory-server/src/test/java/org/opencadc/inventory/transfer/ProtocolsGeneratorTest.java
andamian: OK. Does it mean that it is possible to give a false positive when random randomly returns the same order?

PR #555, opened 2024-02-13
Make `vault` work with existing `vostools`

andamian:


vault/src/main/java/org/opencadc/vault/VaultTransferGenerator.java
```diff
@@ -186,11 +186,9 @@ private List<Protocol> handleDataNode(DataNode node, String filename, Transfer t
         Transfer artifactTrans = new Transfer(node.storageID, trans.getDirection());
         for (Protocol p : trans.getProtocols()) {
             log.debug("requested protocol: " + p);
-            URI secM = p.getSecurityMethod();
-            if (secM == null || secM.equals(Standards.SECURITY_METHOD_ANON)) {
-                artifactTrans.getProtocols().add(p);
-                log.debug("allow protocol: " + p);
-            }
+            p.setSecurityMethod(Standards.SECURITY_METHOD_ANON);
```
pdowler, line 189: won't this add `trans.getProtocols().size()` identical anon protocol requests to the `artifactTrans`? 

what we want is 1 protocol in there with the same protocol uri as the request, without making assumptions about what the protocol actually is. So if you are going to set security method to anon, you need to get the minimal distinct set of protocol uris in the request

also, I would create `new Protocol(...)` and not modify the ones in `trans`

pdowler (changes requested):

vault/src/main/java/org/opencadc/vault/VaultTransferGenerator.java
andamian: Reworked.

pdowler (approved):

PR #557, opened 2024-02-27
cadc-inventory-db update to support node size metadata and workflow

pdowler:


pdowler:
known: CI fail is due to an API change that breaks ringhold

cadc-inventory-db/src/main/resources/vospace.Node.sql
```diff
@@ -19,7 +20,9 @@ create table <schema>.Node (
     
     -- DataNode
     busy boolean,
-    storageID varchar(512),
+    bytesUsed bigint,
```
andamian, line 22: Duplicate with line 9?

vault-quota/Design.md
```diff
@@ -13,6 +13,22 @@ The design below only takes into account incremental propagation of space used
 by stored files. It is not complete/verified until we also come up with a validation
 algorithm that can detect and fix discrepancies in a live `vault`.
 
+## operations that effect node size
```
andamian, line 16: Is this section still required? And the other size related changes?

vault/README.md
```diff
@@ -102,6 +105,15 @@ _all known_ sites. It only makes sense to enable this when `vault` is running in
 `raven` and/or `fenwick` instances syncing artifact metadata. This feature introduces an overhead for the 
 genuine not-found cases: transfer negotiation to GET the file that was never PUT.
 
+The _allocationParent_ is a path to a container node (directory) which contains space allocations. An allocation
+is owned by a user (uisually different from the _rootOwner_ admin user) who is responsible for the allocation
```
andamian, line 109:
```suggestion
is owned by a user (usually different from the _rootOwner_ admin user) who is responsible for the allocation
```

vault/src/main/java/org/opencadc/vault/VaultTransferGenerator.java
```diff
@@ -193,8 +193,8 @@ private List<Protocol> handleDataNode(DataNode node, String filename, Transfer t
             log.debug("requested protocol: " + p);
             if (!protoURIs.contains(p.getUri())) {
                 Protocol anonProto = new Protocol(p.getUri());
-                anonProto.setSecurityMethod(Standards.SECURITY_METHOD_ANON);
-                artifactTrans.getProtocols().add(p);
+                //anonProto.setSecurityMethod(Standards.SECURITY_METHOD_ANON);
```
andamian, line 196: Remove commented out code

vault/src/main/java/org/opencadc/vault/NodePersistenceImpl.java
```diff
@@ -265,6 +297,37 @@ public ContainerNode getRootNode() {
...
     }
 
+    @Override
+    public boolean isAllocation(ContainerNode cn) {
+        if (cn.parent == null) {
+            return false; // root is never an allocation
+        }
+        ContainerNode p = cn.parent;
+        for (ContainerNode ap : allocationParents) {
+            if (absoluteEquals(p.parent, ap)) {
```
andamian: Isn't it sufficient to check the paths in order to conclude that it's the same node, e.g.  `(Utils.getPath(ap).equals(Utils.getPath(p.parent)))`?

andamian (approved):

cadc-inventory-db/src/main/resources/vospace.Node.sql
pdowler: already fixed

vault-quota/Design.md
pdowler: re-working the doc based on our change of plans

vault/README.md
pdowler: was already fixed but not committed

vault/src/main/java/org/opencadc/vault/NodePersistenceImpl.java
pdowler: maybe. I generally don't like relying on string comparisons.

it turns out that in `vault` I could just use the ContainerNode uuid and not look at the path at all. I wrote this code in cavern and copied it, so it works but is more complex than required.

will change to use the uuids and see how it looks.

PR #561, opened 2024-03-14
Data size synchronization worker (CADC 13214)

andamian:


cadc-inventory-db/src/main/java/org/opencadc/inventory/ArtifactSyncWorker.java
```diff
@@ -0,0 +1,143 @@
...
+        try (final ResourceIterator<Artifact> iter = artifactDAO.iterator(storageNamespace, null,
+                harvestState.curLastModified, true)) {
+            while (iter.hasNext()) {
+                Artifact artifact = iter.next();
+                DataNode node = nodeDAO.getDataNode(artifact.getURI());
+                if ((node != null) && !artifact.getContentLength().equals(node.bytesUsed)) {
+                    node.bytesUsed = artifact.getContentLength();
+                    tm.startTransaction();
+                    try {
+                        nodeDAO.put(node);
```
pdowler, line 117: Inside the txn but before the put(node), this should lock the the node before updating in case state changed since the getDataNode call.

The lock method returns another instance of the node that might be different from getDataNode if another process also updated it (eg NodePersistenceImpl.get(...) if we do that "cache update " or validation) and it might return null if the node was deleted... all part of race condition handling (but not easily testable without a lot of effort that's not worth it). Once you have started and txn and locked the node, there is not even much point in rechecking the bytesUsed: just assign the value and put. The metaChecksum will optimize whether the update is needed or not.

-----

for harvestStateDAO.put(...) you do not need the second argument to force the timestamp update. Here it should just rely on changing the fields of the object.

Logically, the harvestStateDAO.put(...) needs to happen for every artifact event handled, not just the ones that cause an update so it needs to go outside the `if (node size needs update)`. That's OK because the harvestStateDAO.put(...) call can happen outside the transaction (it risks re-evaluating an artifact event, which is idempotent so harmless). Because I will probably configure the HarvestStateDAO to buffer updates and possibly even run the maintenance it *must* be outside the transaction because that happens inside the put().

pdowler (changes requested):

pdowler (approved):

PR #564, opened 2024-03-26
CADC-13234 fix ringhold for ArtifactDAO API change

jburke-cadc:


ringhold/README.md
```diff
@@ -1,6 +1,6 @@
-# Storage Inventory local artifact removal process (ringhold)
+# Storage Inventory local artifact deletion process (ringhold)
 
-Process to remove local artifacts that are no longer being synchronised by fenwick. This tool is used
+Process to delete local artifacts that are no longer being synchronised by fenwick. This tool is used
```
pdowler, line 3: Wasn't part of this work specifically, but this change from remove to delete makes it more obvious that this is a little too vague. Maybe something like:
```
Process to remove the local copy of artifacts from a storage site inventory database and generate DeletedStorageLocationEvent(s) so the removal will propagate correctly to a global inventory. This does not remove the files from storage (see `tantar`).
```
I mentioned "storage site" explicitly because this tool doesn't really do the right thing in global (there's a future story for that).

ringhold/README.md
```diff
@@ -18,17 +18,25 @@ org.opencadc.ringhold.inventory.schema={schema for inventory database objects}
 org.opencadc.ringhold.inventory.username={username for inventory admin}
 org.opencadc.ringhold.inventory.password={password for inventory admin}
 org.opencadc.ringhold.inventory.url=jdbc:postgresql://{server}/{database}
+
+# artifact namespace
```
pdowler, line 22: make the comment `artifact namespace(s) to remove`

pdowler (changes requested):

ringhold/README.md
jburke-cadc: Thought we said it should be delete instead of remove, to be more explicit what it is doing., but the above is much better. Updated.

ringhold/README.md
jburke-cadc: Updated.

pdowler (approved):

PR #576, opened 2024-04-10
vault: preventNotFound in files endpoint

pdowler:


vault/src/main/java/org/opencadc/vault/files/HeadAction.java
```diff
@@ -150,29 +163,70 @@ DataNode resolveAndSetMetadata(boolean includeContentLength) throws Exception {
...
+        String clm = node.getPropertyValue(VOS.PROPERTY_URI_CONTENTDATE); // use if present
+        if (clm == null) {
+            clm = node.getPropertyValue(VOS.PROPERTY_URI_DATE);
         }
+        Date contentLastModified = null;
+        if (clm != null) {
+            contentLastModified = NodeWriter.getDateFormat().parse(clm);
+        }
+        if (ret.artifact != null) {
+            // preventNotFound
```
andamian, line 204: This potentially overrides the values set in the previous lines which might not be very obvious. Maybe re-org so that it's more clear where the specific value originates from (artifact first then node). Just a comment.

andamian (approved):

vault/src/main/java/org/opencadc/vault/files/HeadAction.java
pdowler: basically, if preventNotFound logic was triggered above and an artifact was found, it's always right to report that metadata as it's the definitive file metadata

the "else" is potentially getting metadata for an old version of the file, which can't practically be detected

I tried several sequences and this was the most clear to me

PR #583, opened 2024-06-18
Cutout extension fix

at88mph:
Availability checks were broken
`date --utc` doesn't work on MacOS, but `date -u` is universal


at88mph:
Also relies on new versions of `cadc-data-ops-fits` and `nom-tam-fits`.

pdowler (approved):
merging despite CI build fail so I can update other modules to source compat = 11

PR #597, opened 2025-07-07
a collection of minor changes to make it possible to run with minimal AAI

pdowler:
luskan: optional config of allowed users with X509 DN

minoc: optional config of allowed users with X509 DN 
- archiveOperator with read-write
- fileSyncOperator with read-only

raven: optional config to avoid get and put for specific storage sites

tantar: some tweaks to deal with bucket types other than hex

cadc-storage-adapter-fs: complete the LogicalFileSystemStorageAdapter that stores files as specified in the Artifact.uri values -- supports read-only access to the filesystem directly


raven/README.md
```diff
@@ -91,6 +94,11 @@ The _putPreference_ rules are optimizations; they do not restrict the destinatio
 where a namespace is intended to be stored only in some site(s) or where most or all PUTs come from systems that are near one 
 storage site. TODO: support for `GET` preferences? also use client IP address to prioritize?
 
+The _putAvoid_ key configures raven to **never** return PUT URLs to the specified site(s).
+
+The _getAvoid_ key configures raven to **only** return GET URLs to the specified site(s) if the site contains the only copy
```
andamian, line 99: "only copy"? Can it return other sites in this case?

andamian (approved):

cadc-storage-adapter-fs/src/main/java/org/opencadc/inventory/storage/fs/AbstractStorageAdapter.java
```diff
@@ -0,0 +1,758 @@
...
+
+    @Override
+    public Iterator<PutTransaction> transactionIterator() throws StorageEngageException, TransientException {
+        throw new UnsupportedOperationException();
+    }
+    
+    @Override
+    public void get(StorageLocation storageLocation, OutputStream dest)
+        throws ResourceNotFoundException, ReadException, WriteException, StorageEngageException, TransientException {
+        InventoryUtil.assertNotNull(OpaqueFileSystemStorageAdapter.class, "storageLocation", storageLocation);
```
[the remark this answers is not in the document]
pdowler: It probably could, and in fact there is a conditional `if (byteRange != null)` , but the original style was to have methods check/require non-null args (so minoc has to call the right method. The swift adapter has the same style... probably worth noting as technical debt as the code in there grew since the original design.

raven/README.md
pdowler: Yes, the idea is that raven will only output a URL to such sites if the artifact is only available there (that's not exactly _the only copy_ since _available_ also includes transient issues like another site being down... but it minimises the number of GETs to a site

If other sites have an (available) copy, then a `getAvoid` site will not be used

andamian: Thanks. To me the last explanation makes more sense - avoid to GET from this sites unless no other copies exist.

PR #626, opened 2026-04-02
fenwick: experimentalBucketMode to build a new inventory db from scratch

pdowler:


fenwick/src/main/java/org/opencadc/fenwick/ArtifactSync.java
```diff
@@ -142,8 +144,30 @@ static String getHarvestStateName(String instanceName) {
         return instanceName + "/" + Artifact.class.getSimpleName();
     }
 
+    
     @Override
     void doit() throws ResourceNotFoundException, IOException, IllegalStateException, TransientException, InterruptedException {
+        if (experimentalBucketMode) {
+            for (char c1 : InventoryUtil.BUCKET_CHARS.toCharArray()) {
```
andamian, line 151: I'm not sure how to read this. It's also not tested.

fenwick/src/main/java/org/opencadc/fenwick/Main.java
```diff
@@ -163,14 +165,14 @@ public static void main(final String[] args) {
             final ConnectionConfig cc = new ConnectionConfig(null, null, username, password,
                                                              "org.postgresql.Driver", dbUrl);
 
-            final String configuredSQLGenerator = props.getFirstPropertyValue(SQLGENERATOR_CONFIG_KEY);
+            String configuredSQLGenerator = props.getFirstPropertyValue(SQLGENERATOR_CONFIG_KEY);
```
andamian: Why did you remove `final`?

andamian (commented):
I'm not sure I understand the `experimentalMode`. The rest looks good.

fenwick/src/main/java/org/opencadc/fenwick/Main.java
pdowler: I think some use of `final` in there was extraneous and awkward and I probably changed this while trying to make some other config handling consistent. I'll revert this to minimise the diff.

fenwick/src/main/java/org/opencadc/fenwick/ArtifactSync.java
pdowler: Only manually tested at this point; if you think the rest is sound I will add an intTest.

pdowler: A previous change was to have named instances with different filters, so the intent is to have a separate "archive" and "vault" fenwick instance. Currently that's of order 200e6 artifacts each. The standard query to luskan does an "order by lastModified", which is fine in incremental mode but when building a new inventory db from scratch that first query has to sort 200e6 records.

The double loop over the bucket characters (hex chars) makes 2-letter prefixes that are added as a query constraint on uriBucket. This subdivides the sync into 256 small queries instead of 1 large one. It also means there are 256 different HarvestState records (per instance) which need to be manually merged into a single one when returning to incremental mode. I have SQL that can do that but did not implement it into the code. I just hard-coded this initial implementation to 256 buckets to keep it as simple as possible... in future I might make use of multiple buckets with multiple threads so it could be used in regular operation.

fenwick/src/intTest/java/org/opencadc/fenwick/ArtifactSyncTest.java
```diff
@@ -145,16 +152,29 @@ private void doAllArtifacts(boolean track) throws Exception {
...
+                throw ex;
+            }
+        }
         
         // source artifacts
         long t = System.currentTimeMillis();
         final Artifact a1 = new Artifact(URI.create("cadc:TEST/a1"), TestUtil.getRandomMD5(), new Date(t - 300), 8989L);
         final Artifact a2 = new Artifact(URI.create("cadc:TEST/a2"), TestUtil.getRandomMD5(), new Date(t - 200), 8989L);
         final Artifact a3 = new Artifact(URI.create("cadc:TEST/a3"), TestUtil.getRandomMD5(), new Date(t - 100), 8989L);
+        final Set<String> buckets = new TreeSet<>();
```
andamian, line 174: Can you turn this into a map (bucket - last artifact) so you can test a similar test to the non experimental bucket mode? I would also move the `buckets` definition closer to where it's first used (line 280).

andamian (approved):

PR #631, opened 2026-05-14
Git issue: 494 - Add If-Modified-Since support to GetAction & Refactor FitsOperationsTest

aratikakadiya:
- GetAction: handle `If-Modified-Since` request header; return HTTP 304 Not Modified when the artifact's contentLastModified is not after the date mentioned in this header, skipping the data transfer entirely
- Added integration test for the new header behaviour verification.
- Updated FitsOperationsTest: reuse the existing artifact URI/URL instead of creating a separate artifact for the subtests.


minoc/src/main/java/org/opencadc/minoc/GetAction.java
```diff
@@ -215,6 +217,21 @@ public void doAction() throws Exception {
             
             // default: complete download
             HeadAction.setHeaders(artifact, filenameOverride, syncOutput);
+
+            String ifModifiedSince = syncInput.getHeader("If-Modified-Since");
+            if (ifModifiedSince != null && artifact.getContentLastModified() != null) {
+                try {
+                    DateFormat df = DateUtil.getDateFormat(DateUtil.HTTP_DATE_FORMAT, DateUtil.UTC);
```
pdowler: In the HTTP spec, the Last-Modified header that `minoc` sends (HEAD and GET) is in the a specific format and is GMT rather than UTC (thanks Sir TBL). See cadc-rest SyncOutput.setLastModified().

Sir TBL = Sir Tim Berners-Lee

pdowler (changes requested):
API doc update: nice

other fixes: good

minoc/src/intTest/java/org/opencadc/minoc/GetActionTest.java
```diff
@@ -0,0 +1,168 @@
...
+                    long contentLength = get.getContentLength();
+                    Assert.assertEquals(computeChecksumURI(data1.getBytes()), digest);
+                    Assert.assertEquals(data1.getBytes().length, contentLength);
+
+                    Thread.sleep(5000); // wait 2 seconds
+
+                    // 304 Not Modified
+                    get = new HttpGet(artifactURL, out);
+                    DateFormat df = DateUtil.getDateFormat(DateUtil.HTTP_DATE_FORMAT, DateUtil.GMT);
+                    String httpDate = df.format(new Date());
```
pdowler: this test will work reliably when the service is running locally (eg minoc and test code using same clock), but might not if a developer runs containers on a different machine

more importantly, the sleep above means this only tests the "greater than" in the date comparison and not the equals... a client that downloads a file will likely store it and assign the timestamp from the Last-Modified header (aka Artifact.contentLastModified) so that's what the conditional get would use and that specific date hits the "greater than or equal to" detail... 

and then the other subtle bit is that HTTP date format only has seconds so that date is likely truncated (would need to look at HeadAction to see how that output works) so it's possible that truncation and "or equal to" only matters 1 in 1000 #ouch

anyway, I think the test is ever so slightly better of httpDate comes from above (header from the previous HttpGet) and then no sleep is needed.

pdowler (changes requested):
One detail noted in the test code.

The other thing that I thought I posted before but maybe did not: it would be better if the new test method was added to BasicOpsTest and we didn't have a GetActionTest. Test code can never really target one action and usually needs several, hence thse are typically a sequence of calls with the full lifecycle.

So, for now just move the new test method to BasicOpsTest and remove this file. With more effort the 304 test could be slid into the existing test code, but this is fine, maybe better.

pdowler (approved):

PR #633, opened 2026-05-25
Git issue: 614 - vault: copy and store additional Artifact props

aratikakadiya:
- Persist VOSpace content metadata properties (#content-date and #content-md5) to DataNode properties
  - Instead of adding them only to the GET response object
  - During the Data Node size sync process
- Update DataNodeSizeWorkerTest.java to validate persisted metadata changes
- Add a test to the NodesTest.java to validate the GET action with persisted content metadata properties

### DIFF
--- a/cadc-inventory-db/src/main/java/org/opencadc/vospace/db/DataNodeSizeWorker.java
+++ b/cadc-inventory-db/src/main/java/org/opencadc/vospace/db/DataNodeSizeWorker.java
@@ -159,9 +159,9 @@ public class DataNodeSizeWorker implements Runnable {
                 if (node != null) {
                     NodeProperty contentChecksumProp = node.getProperty(VOS.PROPERTY_URI_CONTENTMD5);
                     boolean isMd5 = "md5".equalsIgnoreCase(artifact.getContentChecksum().getScheme());
-                    boolean updateContentChecksum = contentChecksumProp == null
-                            || !isMd5 // need to remove existing property if checksum is no longer MD5
-                            || !artifact.getContentChecksum().getSchemeSpecificPart().equals(contentChecksumProp.getValue());
+                    boolean updateContentChecksum = (contentChecksumProp == null && isMd5) // need to create new property for checksums
+                            || (contentChecksumProp != null // need to remove existing property if checksum is no longer MD5
+                                && (!isMd5 || !artifact.getContentChecksum().getSchemeSpecificPart().equals(contentChecksumProp.getValue())));
 
                     NodeProperty contentDateProp = node.getProperty(VOS.PROPERTY_URI_CONTENTDATE);
                     String contentLastModifiedStr = df.format(artifact.getContentLastModified());
--- a/cadc-inventory-db/src/main/java/org/opencadc/vospace/db/DataNodeSizeWorker.java
+++ b/cadc-inventory-db/src/main/java/org/opencadc/vospace/db/DataNodeSizeWorker.java
@@ -158,10 +158,10 @@ public class DataNodeSizeWorker implements Runnable {
                 DataNode node = nodeDAO.getDataNode(artifact.getURI());
                 if (node != null) {
                     NodeProperty contentChecksumProp = node.getProperty(VOS.PROPERTY_URI_CONTENTMD5);
+                    boolean isMd5 = "md5".equalsIgnoreCase(artifact.getContentChecksum().getScheme());
                     boolean updateContentChecksum = contentChecksumProp == null
-                            || !artifact.getContentChecksum().getScheme().equalsIgnoreCase("md5") // remove existing property if checksum is no longer MD5
-                            || (artifact.getContentChecksum().getScheme().equalsIgnoreCase("md5") // Persist VOSpace content-md5 property only for MD5 checksums
-                                    && !artifact.getContentChecksum().getSchemeSpecificPart().equals(contentChecksumProp.getValue()));
+                            || !isMd5 // need to remove existing property if checksum is no longer MD5
+                            || !artifact.getContentChecksum().getSchemeSpecificPart().equals(contentChecksumProp.getValue());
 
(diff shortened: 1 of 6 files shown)


cadc-inventory-db/src/main/java/org/opencadc/vospace/db/DataNodeSizeWorker.java
```diff
@@ -165,6 +167,19 @@ public void run() {
                             continue; // node gone - race condition
                         }
                         node.bytesUsed = artifact.getContentLength();
+
+                        // Persist #content-date only when it differs from the node #date value
```
pdowler: We must always store the contentLastModified property because the Node.lastModified timestamp can change in the future (due to any mutable change). It is only on output that we don't output both if they are equal. So: always add the CONTENTDATE prop.

The check on "md5" algorithm is a good detail; SI can in theory support other algorithms (we just don't) but VOSpace standard is more limited.

note: Node.properties is a Set so yes, you would have to remove before add. That requires two traversals of the set, but the set of props is usually quite small, so that's probably fine. Alternatively, the NodeProperty value is mutable, so you could find the existing property by key and set the value and only add if it was missing. In principle, a single set traversal could be faster when the set is large enough and I suspect `removeIf` iterates... but it is fine for now.

vault/src/main/java/org/opencadc/vault/NodePersistenceImpl.java
```diff
@@ -416,6 +416,19 @@ public Node get(ContainerNode parent, String name) throws TransientException {
                         if (locked != null) {
                             dn = locked; // safer than accidentally using the wrong variable
                             dn.bytesUsed = a.getContentLength();
+
+                            // Persist #content-date only when it differs from the node #date value
+                            if (!ret.getLastModified().equals(a.getContentLastModified())) {
```
pdowler: same comment here as in DataNodeSizeWorker

pdowler (changes requested):
main code review (haven't looked at test code yet)

pdowler (changes requested):
There is still a subtle issue in the DataNodeSizeWorker that I missed earlier.

At lines 160-161 is some logic to detect if there is actually a `delta` to apply and it looked for a size change. I am sure I wrote it but I don't know why it also compares lastModified timestamps... I might have been future-proofing this for any other cases where the artifact changed after the node.

The thing is that many files can end up being the same size when recreated by processing (eg FITS files are always a multiple of 2880 byte blocks w/ padding so they have to change size by more than one block for file size to actually change).

In SI, Artifact.contentLength, contentChecksum, and contentLastModified are immutable: the values are assigned at the end of a PUT and cannot change. When a file is overwritten in SI, that is actually a new Artifact and a new DeletedArtifactEvent (for the old one).

So, given that, looking for a change in size seems like a subtle optimisation. The correct `delta` would compare all 3 of the values we are copying to the DataNode. If we want to optimise/simplify, we could just compare the Artifact.contentLastModified (because someone could, in principle, overwrite a file with an identical copy and thus only change that timestamp) as that is the one prop that always changes... but if we can make the code clean and simple to read it would be best to compare all 3.

The comparison of Artifact.lastModified to Node.lastModified would detect most cases where the file in SI changed, but it could also miss some if some other aspect of the node was updated (like other node properties)... so that might even now be a subtle bug.

cadc-inventory-db/src/main/java/org/opencadc/vospace/db/DataNodeSizeWorker.java
```diff
@@ -155,8 +157,17 @@ public void run() {
                 Artifact artifact = iter.next();
                 DataNode node = nodeDAO.getDataNode(artifact.getURI());
                 if (node != null) {
-                    boolean delta = !artifact.getContentLength().equals(node.bytesUsed); // size change
-                    delta = delta || artifact.getLastModified().after(node.getLastModified());
+                    NodeProperty contentChecksumProp = node.getProperty(VOS.PROPERTY_URI_CONTENTMD5);
+                    boolean updateContentChecksum = contentChecksumProp == null
+                            || (artifact.getContentChecksum().getScheme().equalsIgnoreCase("md5") // Persist VOSpace content-md5 property only for MD5 checksums
```
pdowler: If the existing prop is null (eg a new DataNode and Artifact) then the md5 check is skipped... then down on line 190 we could (in principle) blindly put a sha1 value in that property :-) 

Also, if the current value is set and the new one is a different algorithm, we would want to clear the property and not insert the new one.

So: the update below needs to recheck the contentChecksum scheme before adding the new value.

cadc-inventory-db/src/main/java/org/opencadc/vospace/db/DataNodeSizeWorker.java
```diff
@@ -155,8 +157,17 @@ public void run() {
...
+                                    && !artifact.getContentChecksum().getSchemeSpecificPart().equals(contentChecksumProp.getValue()));
+
+                    NodeProperty contentDateProp = node.getProperty(VOS.PROPERTY_URI_CONTENTDATE);
+                    boolean updateContentDate = contentDateProp == null || !df.format(artifact.getContentLastModified()).equals(contentDateProp.getValue());
```
pdowler: I would have used `df.parse(contentDateProp)` and then compare Date objects, but your approach is closer to not having to parse and/or format the twice. 

So: capture the formatted value in a local var, compare the strings, and re-use the formatted value below when adding/replacing the prop.

pdowler (changes requested):

pdowler (dismissed):

pdowler (approved):

pdowler (dismissed):

pdowler (changes requested):
two final changes to do:
- increment the version of cadc-inventory-db to 1.1.1
- increment the version of vault to 1.1.2

pdowler (approved):

